PR #46, opened 2019-01-17
Pass on the hw button keycodes

### LINKED ISSUE #45: Route KEY_PROG1 and friends through to X
The Cintiq 24HD sends `KEY_PROG1` through to `KEY_PROG3` for the special buttons, but the wacom driver discards them. Let's not do that, let's pass them on and let userspace figure out what to do with them. The default mapping is `XF86Launch1` through to `XF86Launch3` so they could be assigned.

More recent tablets send `KEY_CONTROLPANEL`, `KEY_ONSCREEN_KEYBOARD`, and `KEY_BUTTONCONFIG` but those have keycodes outside the 255 X keycode range.

whot:
Fixes #45 

**This is still work in progress**

This patch adds the code to pass on `KEY_PROG1` and friends to the X server, to be passed on to the clients from there. Realistically, this doesn't do anything for the other three codes because they're above the 255 keycode limit and will get dropped by the server immediately.

The handling is hardcoded for the 6 buttons we care about, IMO there's little point to make this a fully generic.

I'm not sure if the events can occur on their own without any `EV_ABS` like we do for the normal buttons. Right now, the code assumes they are not affected by proximity, etc.

cc @jigpu, @Pinglinux, @skomra

### DIFF
--- a/src/wcmCommon.c
+++ b/src/wcmCommon.c
@@ -195,6 +195,47 @@ static void wcmSendButtons(InputInfoPtr pInfo, const WacomDeviceState* ds, int b
 
 }
 
+static void wcmSendKeys (InputInfoPtr pInfo, unsigned int current, unsigned int previous)
+{
+	unsigned int mask, idx;
+	WacomDevicePtr priv = (WacomDevicePtr) pInfo->private;
+
+	DBG(6, priv, "current=%d previous=%d\n", current, previous);
+
+	for (idx = 0, mask = 0x1;
+	     mask && (mask <= current || mask <= previous);
+	     idx++, mask <<= 1)
+	{
+		if ((mask & previous) != (mask & current))
+		{
+			int key = 0;
+			int state = !!(mask & current);
+
+			switch (idx) {
+				/* Note: the evdev keycodes are > 255 and
+				 * get dropped by the server. So let's remap
+				 * those to KEY_PROG1-3 instead */
+				case IDX_KEY_CONTROLPANEL:
+					key = KEY_PROG1;
+					break;
+				case IDX_KEY_ONSCREEN_KEYBOARD:
+					key = KEY_PROG2;
+					break;
+				case IDX_KEY_BUTTONCONFIG:
+					key = KEY_PROG3;
+					break;
+				case IDX_KEY_INFO:
+					key = KEY_PROG4;
+					break;
+				default:
+					break;
+			}
+			if (key)
+				wcmEmitKeycode(pInfo->dev, key + 8, state);
+		}
+	}
+}
+
 void wcmEmitKeycode (DeviceIntPtr keydev, int keycode, int state)
 {
 	xf86PostKeyboardEvent (keydev, keycode, state);
@@ -600,6 +641,8 @@ wcmSendPadEvents(InputInfoPtr pInfo, const WacomDeviceState* ds,
 			wcmSendButtons(pInfo, ds, ds->buttons, first_val, num_vals, valuators);
 	}
 
+	wcmSendKeys(pInfo, ds->keys, priv->oldState.keys);
+
 	if (priv->oldState.proximity && !ds->proximity)
 		xf86PostProximityEventP(pInfo->dev, 0, first_val, num_vals,
 					VCOPY(valuators, num_vals));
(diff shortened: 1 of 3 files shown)


whot:
tbh, I doubt you'd get that change through. You'd be breaking userspace by changing the button codes.

jigpu:
I kinda doubt there's any userspace code *to* be broken. Definitely not xf86-input-wacom or libinput, and a vanishingly small chance of some custom script or program going to the trouble of watching the evdev device to read the state of these buttons. If its a concern though, we could continue posting the KEY_PROG* events and just post the "proper" events alongside them.

whot:
Well, from my POV anything you do to make the kernel report saner events is going to be welcomed. I'd be ok with posting both as an emergency case, I can disable/ignore those quite easily.

So if you submit a patch for that, just picture me with pom-poms on the side, cheering you on. Or not, if not picturing that is more helpful ;)

skomra:
Kernel patch queued for 5.4: https://patchwork.kernel.org/patch/11098125/

jigpu:
The kernel changes appear to be in place, so so we want to remove the references to `KEY_PROG` references from this patch and then consider merging it?

whot:
Well... `KEY_PROG` were the ones below the 255 keycode limit, if we remove the processing of those keys then we just end up passing events to the server which will immediately drop them. The chances of X ever having 32-bit keycodes are nil at this point, so adding the patch will merely help with keeping the room warm but not add any functionality.

whot:
Rebased, and I've dropped support for `KEY_PROG1-3` now. To make the `KEY_BUTTONCONFIG` and friends work, I had to remap those to ... `KEY_PROG1` and friends. 255 keycode limit and whatnot. Alanis Morisette may have something to say about this.

If merged, we can add an xkeyboard-config option to map those to some keysyms, but for now this here will do.

jigpu:
@whot main obvious concern is missing support for `KEY_INFO` (see suggested patch below). Other than that, it looks fine to me. Is there any reason I shouldn't merge this into the release I'm about to make?

~~~
diff --git a/src/wcmCommon.c b/src/wcmCommon.c
index a55c956..8f563ac 100644
--- a/src/wcmCommon.c
+++ b/src/wcmCommon.c
@@ -224,6 +224,9 @@ static void wcmSendKeys (InputInfoPtr pInfo, int current, int previous)
 				case IDX_KEY_BUTTONCONFIG:
 					key = KEY_PROG3;
 					break;
+				case IDX_KEY_INFO:
+					key = KEY_PROG4;
+					break;
 				default:
 					break;
 			}
diff --git a/src/wcmUSB.c b/src/wcmUSB.c
index 25e2897..d0220af 100644
--- a/src/wcmUSB.c
+++ b/src/wcmUSB.c
@@ -1647,6 +1647,9 @@ static void usbParseBTNEvent(WacomCommonPtr common,
 		case KEY_BUTTONCONFIG:
 			ds->keys = mod_buttons(ds->keys, IDX_KEY_BUTTONCONFIG, event->value);
 			break;
+		case KEY_INFO:
+			ds->keys = mod_buttons(ds->keys, IDX_KEY_INFO, event->value);
+			break;
 		default:
 			for (nkeys = 0; nkeys < usbdata->npadkeys; nkeys++)
 			{
diff --git a/src/xf86WacomDefs.h b/src/xf86WacomDefs.h
index 6b1744f..164509a 100644
--- a/src/xf86WacomDefs.h
+++ b/src/xf86WacomDefs.h
@@ -213,6 +213,7 @@ struct _WacomModel
 #define IDX_KEY_CONTROLPANEL		1
 #define IDX_KEY_ONSCREEN_KEYBOARD	2
 #define IDX_KEY_BUTTONCONFIG		3
+#define IDX_KEY_INFO			4
 
 /******************************************************************************
  * WacomDeviceState
~~~

whot:
Thanks, I've squashed that diff in (together with a few minor changes to the existing code etc). I think it's ready to merge but I haven't tested this since the last attempt (Sep. If you can fit a test in please do merge, otherwise better to hold off until I find time to do that.

jigpu:
Quick test with `xinput test-xi2` showed the scancodes coming through as expected, so I've gone ahead and merged.

Lovewyrm:
Just to ask, as a layman.
Does this mean it's possible to soon/now bind actions to these keys?

whot:
Yes, they should have keycodes 156, 157 and 210 (`xev` will verify this). These should be mapped into keycodes `XF86Launch1` and so on so you can work from that.

whot:
> I have a cintiq pro 24 and the only way I can even see the wacom devices is by running evemu as root.

that is expected, input devices cannot usually be opened by the user as this leads to keyloggers. What events do you see when you press the buttons?

whot:
*could* be that your desktop environment grabs those calls to launch the matching program but then doesn't actually launch anything (maybe because there is nothing configured).  Or could be that the driver has a bug, but a quick check shows that the code to convert those into `KEY_PROG1` etc is still there and that those are mapped into `XF86Launch1`  in xkeyboard-config.

The former can be ruled out if you start a plain X session (see `xinit`) and then run `xev -root` - if that doesn't show any events either the issue is within the driver or the server (though the server *shouldn't* care about this at all).

PR #54, opened 2019-04-12
Release script fixes

whot:


whot:
> This looks good to me, I did not test as testing without actually releasing is quite a pain.

`--dry-run` should work correctly now. So you can git clone it into a tmp directory, add a fake tag and then do a test with `--dry-run`. But yeah, something will always be missing. Thanks for the review.

> The release documentation[1] will need to be updated.

Done, I think :)

Issue #52, opened 2019-04-08
Surface Go: pen triggers eraser under X11 but not Wayland


ShapeShifter499:
After some back and forth on this https://github.com/linuxwacom/libwacom/issues/75  It was determined the issue lies somewhere in xf86-input-wacom. But the exact fix is unknown and it's not clear what part of code is failing here.

When using the pen in X11 and Gnome, the pen is triggering the eraser (or whatever tool the 'eraser' end was set to). Applications like GIMP, Xournal, and Stylus Labs Write all showed the same issue. 

However under Wayland and Gnome these issues went away. The pen tip wrote fine with pressure sensitivity and the eraser end erased lines. All previously tested applications no longer had issue.

So this is where I'm at now, I'd like to see X11 support fixed as there are still some quirks with Wayland.

Baldurius:
I have the same problems as @ShapeShifter499 .

I've compiled the wacom driver following the [instructions](https://github.com/linuxwacom/xf86-input-wacom/wiki/Building-The-Driver) looked into the log files by first running
```
xsetwacom list
```
which printed
```
ELAN9038:00 04F3:261A stylus    id: 16   type: STYLUS
ELAN9038:00 04F3:261A eraser    id: 17   type: ERASER
```
I then enable debug messages via
```
xsetwacom --set "ELAN9038:00 04F3:261A stylus" "ToolDebugLevel" 12
xsetwacom --set "ELAN9038:00 04F3:261A eraser" "ToolDebugLevel" 12
```
and looked into `/var/log/Xorg.0.log`.

Here is one moment where the stylus switches to the eraser:
```
[  1555.074] (II) ELAN9038:00 04F3:261A stylus (7:wcmSendEvents): [STYLUS] o_prox=true x=7210 y=3956 z=1567 b=true b=1 tx=10 ty=-26 wl=0 wl2=0 rot=0 th=0
[  1555.074] (II) ELAN9038:00 04F3:261A stylus (10:wcmRotateAndScaleCoordinates): rotate/scaled to 7210/3956
[  1555.074] (II) ELAN9038:00 04F3:261A stylus (6:wcmSendEvents): abs prox=1	x=7210	y=3956	z=1567	v3=10	v4=-26	v5=-900	v6=0	id=4500	serial=1	button=true	buttons=1
[  1555.074] (II) ELAN9038:00 04F3:261A eraser (10:wcmReady): 0 numbers of data
[...]

PR #62, opened 2019-06-10
Fix issues with Microsoft Surface Go tablet

jigpu:
The Surface Go sends `ABS_MISC` events from the kernel that confuse our driver. This pull request ensures that the generic protocol is always used for non-Wacom devices and that the driver properly handles usually-non-generic data from a generic tablet.

Ref: https://github.com/linuxwacom/xf86-input-wacom/issues/52/


Pinglinux:
Patchset looks good.

Issue #86, opened 2019-10-08
Wacom crashes after suspend on X200T


Akraboy:
I have a ThinkPad X200 Tablet running Arch Linux (5.3.5) and when I suspend the device the touch and pen inputs stop working.
I think my problem is different from https://github.com/linuxwacom/input-wacom/issues/110 so I'm making a new issue.
I also tried reloading wacom and wacom_w8001 but that doesn't fix it. Rebooting however does. It doesn't have to be a cold boot. My journal-log shows that isdv4-serial-inputattach crashed.

Journalctl:

[journal.log](https://github.com/linuxwacom/input-wacom/files/3702028/journal4.log)

Dmesg:

[dmesg.log](https://github.com/linuxwacom/input-wacom/files/3702029/dmesg4.log)

edit: I just messed around with suspend again and had touch working a few times after waking up the device so this doesn't happen 100% of the time.

jigpu:
It looks like the crash is caused by some function writing past the end of a buffer. Unfortunately the stack trace contained in your journalctl doesn't provide the name of the misbehaving function. If you are able to occasionally reproduce this, I would suggest building and installing a version of the driver with debug information included. Since you're using Arch, you should be able to do that by following these steps:

1. Create a temporary directory and save the PKGBUILD from https://git.archlinux.org/svntogit/packages.git/plain/trunk/PKGBUILD?h=packages/xf86-input-wacom into it
2. Open a terminal and navigate to the temporary directory
3. Run `sed -i -e '/conflicts=/a\' -e 'options=(debug !strip)' PKGBUILD` to ensure the package is built with the necessary debug flags
4. Run `sudo pacman -S base-devel` to install the necessary development tools
5. From inside the temporary directory run `makepkg -s -f` to install remaining build dependencies, compile the driver, and create a package
6. Run `sudo pacman -U xf86-input-wacom-0.37.0-1-x86_64.pkg.*` to install the package
7. Reboot

Afterwards if you are able to trigger the issue again, run `sudo coredumpctl info isdv4-serial-inputattach` to get a copy of the relevant debug info for us.

Akraboy:
`sudo coredumpctl info isdv4-serial-inputattach` returns: 
>No coredumps found.

`sudo coredumpctl info isdv4-serial-in` works though and it returned this:

[isdv4-serial-in.txt](https://github.com/linuxwacom/xf86-input-wacom/files/3733682/isdv4-serial-in.txt)

[...]

PR #90, opened 2019-10-18
tools: Fix potential buffer overflow when reading from serial tablet

jigpu:
The read_data() function is given a "min_len" number of bytes to read
to ensure that a complete data structure is read, regardless of garbage
that may be on the line. When garbage is present, however, it can
potentially overflow the buffer.

The function already has code to memmove the good data over garbage and
perform re-reads until "min_len" bytes of good data are available. All
we need to do to avoid the buffer overflow is ensure that the maximum
number of bytes we read() in one call is no more than the number of
bytes free at the end of the buffer.

Ref: https://github.com/linuxwacom/xf86-input-wacom/issues/86
Fixes: 3546d8ab1b ("tools: add isdv4-serial-debugger test program")

### DIFF
--- a/tools/tools-shared.c
+++ b/tools/tools-shared.c
@@ -219,7 +219,7 @@ int read_data(int fd, unsigned char* buffer, int min_len)
 	TRACE("Reading %d bytes from device.\n", min_len);
 redo:
 	do {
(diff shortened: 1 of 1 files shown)


whot:
LGTM

Issue #74, opened 2019-08-15
pen/eraser confusion with wacom bamboo ink plus


cstoitner:
On my Lenovo X1 Tablet 3. gen the Wacom bamboo ink plus stylus seems to confuse eraser and pen mode. Usually the eraser-tool is selected when the stylus is brought near the display, irrespective of pushing the eraser button. Sometimes the normal pen tool is selected instead. When the eraser button is pushed after they stylus is already near the display then the pen tool is usually selected (and it stays that way even after the button is released).

Kritas tablet tester windows usually prints the following when the stylus is brought near the display and than taken away again:
```
Pen tip taken away
Eraser brought near
...
Eraser taken away
Pen tip taken away
```
With the Lenovo pen it looks like this instead:
´´´
Pen tip brought near
...
Pen tip taken away
´´´

The Lenovo pen included with the tablet works without problems. Both pens work on windows without problems. I am running debian bullseye with xserver-xorg-input-wacom 0.34.99.1-1.

This bug sound pretty similar to https://github.com/linuxwacom/xf86-input-wacom/issues/39 and maybe https://github.com/linuxwacom/xf86-input-wacom/issues/52.

I have run the `capture.sh` script from the other issue on the stylus device, drawing a short line with either pen without any buttons pressed.
[lenovo_pen_stylus.tar.gz](https://github.com/linuxwacom/xf86-input-wacom/files/3506735/lenovo_pen_stylus.tar.gz)
[wacom_bamboo_ink_plus_stylus.tar.gz](https://github.com/linuxwacom/xf86-input-wacom/files/3506736/wacom_bamboo_ink_plus_stylus.tar.gz)

cstoitner:
The Wacom pen seems to work correctly with the libinput driver, though the Lenovo pen lost the eraser function.

jigpu:
I've replayed your logs locally and don't see the problem with xf86-input-wacom 0.37.0. This could mean that the problem is fixed with the latest version (e.g. by the fix for #39). Please try installing the driver from source and see if that helps. Instructions can be found [here](https://github.com/linuxwacom/xf86-input-wacom/wiki/Building-The-Driver).

jigpu:
Oops, looks like I tested the wrong script :flushed: 

[...]

PR #92, opened 2019-11-01
Ignore ABS_MISC as a source of device type information for AES pens

jigpu:
AES sensors use protocol 5 since they send ABS_MISC events which contain
information about the tool type in use. The tool type information sent
by AES sensors does not match that used by EMR sensors, however. In
particular, it is not possible to extract stylus/eraser/puck information
from the ID.

The driver would normally never try to extract this information, but the
problem was highlighed when a bug in the kernel would cause the device ID
to be reported twice: once in a packet alongside a BTN_TOOL_* event (fine)
and a second time in a packet without such an event (causing the driver
to try to figure it out from the ID instead).

This commit adds detection for AES pen IDs and does not try to extract
such information if an AES pen is in use.

Ref: https://github.com/linuxwacom/input-wacom/issues/134
Fixes: https://github.com/linuxwacom/xf86-input-wacom/issues/74

### DIFF
--- a/src/wcmUSB.c
+++ b/src/wcmUSB.c
@@ -464,6 +464,21 @@ static Bool usbWcmInit(InputInfoPtr pInfo, char* id, size_t id_len, float *versi
 		common->wcmResolX = common->wcmResolY = 1016;
 	}
 
+	/* Mark protocol 5 devices which predate the use of Intuos5 tech
+	 * as using legacy IDs. These legacy IDs do not conform to the
+	 * current format, in particular with respect to the use of the
+	 * 0th device ID bit signalling EMR/AES.
+	 */
+	if (common->wcmModel == &usbIntuos || common->wcmModel == &usbIntuos2 ||
+	    common->wcmModel == &usbIntuos3 || common->wcmModel == &usbIntuos4 ||
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0x3F) || // Cintiq 21UX (Intuos3-era)
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0xC5) || // Cintiq 20WSX (Intuos3-era)
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0xC6) || // Cintiq 12WX (Intuos3-era)
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0xCC))   // Cintiq 21UX2 (Intuos4-era)
+	{
+		TabletSetFeature(common, WCM_LEGACY_IDS);
+	}
+
 	/* Find out supported button codes. */
 	usbdata->npadkeys = 0;
 	for (i = 0; i < ARRAY_SIZE(padkey_codes); i++)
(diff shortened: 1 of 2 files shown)


src/wcmUSB.c
```diff
@@ -464,6 +464,21 @@ static Bool usbWcmInit(InputInfoPtr pInfo, char* id, size_t id_len, float *versi
...
+	 * as using legacy IDs. These legacy IDs do not conform to the
+	 * current format, in particular with respect to the use of the
+	 * 0th device ID bit signalling EMR/AES.
+	 */
+	if (common->wcmModel == &usbIntuos || common->wcmModel == &usbIntuos2 ||
+	    common->wcmModel == &usbIntuos3 || common->wcmModel == &usbIntuos4 ||
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0x3F) ||
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0xC5) ||
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0xC6) ||
+	    (sID.vendor == WACOM_VENDOR_ID && sID.product == 0xCC))
```
whot: I'd really prefer the ids to be expanded to identifiable names here. Or at least some comment because it's impossible to check whether the code is correct without any intention. Could be a typo in there but we'd never know.

3f is the 21UX, c5 the 20WSX? and so on?

whot:
I'm not 100% clear on the conditions, especially since I may need something similar in libinput. Any extra info would be appreciated:
- libinput merely passes the tool id on to the caller but doesn't use it internally. so I don't think I'll need changes here, right?
- libinput uses the `BTN_TOOL_*` to determine the actual tool. Why do we do the tool ID dance in this driver here instead of just punting this to the kernel and rely on the `BTN_TOOL_*` bits instead? is libinput buggy here or is the driver just still doing something for historical reasons that's no longer required?

jigpu:
> * libinput merely passes the tool id on to the caller but doesn't use it internally. so I don't think I'll need changes here, right?

Correct. You'd only care if you're teasing apart the bits for some reason.

> * libinput uses the `BTN_TOOL_*` to determine the actual tool. Why do we do the tool ID dance in this driver here instead of just punting this to the kernel and rely on the `BTN_TOOL_*` bits instead? is libinput buggy here or is the driver just still doing something for historical reasons that's no longer required?

We use the `BTN_TOOL_*` events as they are sent by the kernel in almost all cases, but this code is part of a fallback path that I believe primarily exists to handle cases where the tool is in prox before the driver starts (e.g. a puck resting on the tablet). The fallback path used to just rely on the `EVIOCGKEY` ioctl to get the current state if it was missing, but commit e08c45f3caf371a8e4123119aeb0b0d7fc227cca added this preferential use of `ABS_MISC` to address some issue with stale type information that I don't understand very well.

I suspect the code isn't necessary with newer kernels that have the stylus and pad devices split apart, but pre-3.17 kernels could still potentially need it.

src/wcmUSB.c
jigpu: Comments added to make the intended devices more explicit.

whot:
thanks for the info, much appreciated. I admit I don't fully understand the motivation behind e08c45f but I think at this point it's old enough that I probably don't have to care about it anymore ;)

PR #129, opened 2020-06-05
Better explanation of the "ToolSerials" option

### LINKED ISSUE #128: [Documentation/Feature request] Guidance on configuring additional input devices
I have an Intuos3 with some of the optional accessories. I cannot find any documentation on how to configure additional styli as input devices (e.g. airbrush, marker pen). xsetwacom only lists the defaults:
```
Wacom Intuos3 9x12 Pad pad              id: 17  type: PAD       
Wacom Intuos3 9x12 Pen stylus           id: 18  type: STYLUS    
Wacom Intuos3 9x12 Pen eraser           id: 19  type: ERASER    
Wacom Intuos3 9x12 Pen cursor           id: 20  type: CURSOR    
```
evtest shows that additional tool options are supported (and the airbrush code is fired when I bring it in proximity. Similarly, the art pen does work with rotation if one configures the dynamics correctly.

```
    Event code 320 (BTN_TOOL_PEN)
    Event code 321 (BTN_TOOL_RUBBER)
    Event code 322 (BTN_TOOL_BRUSH)
    Event code 323 (BTN_TOOL_PENCIL)
    Event code 324 (BTN_TOOL_AIRBRUSH)
    Event code 326 (BTN_TOOL_MOUSE)
    Event code 327 (BTN_TOOL_LENS)
```

Perhaps I am misunderstanding how this should work, but right now, I only get two tool types in Gimp - stylus and eraser. I'm trying to find out how to configure it such that I can have each stylus option have its own remembered tool - for example, if I bring the airbrush over the pad, it changes to that tool, if I bring the eraser or pen in proximity, it changes to those (To clarify - I know gimp remembers the last thing you've selected with each input device and that functionality is not your responsibility - I am just looking for a way to add more "extended input devices" for gimp to use). 

I have the following package versions installed on OpenSuSE 42.3, and the data files look to contain the needed information/ID references for all of the tools.
xf86-input-wacom-0.32.0-4.5.x86_64
libwacom2-0.23-2.1.x86_64
libwacom-data-0.23-2.1.x86_64
libwacom-tools-0.23-lp150.1.9.x86_64

```
[Device]
Name=Wacom Intuos3 9x12
DeviceMatch=usb:056a:00b2
Class=Intuos3
Width=12
Height=9
Layout=intuos3-9x12.svg
Styli=0x913;0x91b;0x813;0x81b;0x823;0x82b;0x801;0x885;0x017;0x097;
```

Additionally, is it possible to reconfigure the input type? For example, if I have both a grip pen and a classic pen, can I set it up so the classic pen registers as `BTN_TOOL_PENCIL` instead? Icing on the cake would be the ability to configure erasers on a per-tool basis, since they do have different tool IDs

Thank you! - I am happy to help provide debugging or testing as necessary.

vintagepc:
Closes #128 

This adds a better explanation of the "ToolSerials" option and why someone might want to use it - this was not readily apparent, and it is very useful for tablets that support the feature when you have multiple physical styli

### DIFF
--- a/man/wacom.man
+++ b/man/wacom.man
@@ -235,10 +235,10 @@ sets the list of serial numbered devices that need to be hotplugged for a physic
 device. The 'type' option may be any of "pen", "airbrush", "artpen", or "cursor".
 This option is only available on wacom V devices (Intuos series and Cintiq 21U).
 To see which serial number belongs to a device, you need to run the utility program,
-xsetwacom, that comes with this driver. In layman's terms, this option will add X
-(virtual) slave pointers for this specific tool's tip (and eraser, if applicable).
+xsetwacom, that comes with this driver. In layman's terms, this option will add 
+additional X devices for this specific tool's tip (and eraser, if applicable).
 This is useful for programs like gimp (which remembers tools based on X device) to recall 
-additional drawing tool selections for your airbrush+eraser, art pen, extra pen, etc. 
+additional drawing tool selections for an airbrush+eraser, art pen, extra pen, etc. 
 .TP 4
 .B Option \fI"Threshold"\fP \fI"number"\fP
 sets the pressure threshold used to generate a button 1 events of stylus.
--- a/man/wacom.man
+++ b/man/wacom.man
@@ -233,8 +233,9 @@ xsetwacom, which comes with linuxwacom package.
 .B Option \fI"ToolSerials"\fP \fI"number[,type[,label]][;...]"\fP
 sets the list of serial numbered devices that need to be hotplugged for a physical
 device. The 'type' option may be any of "pen", "airbrush", "artpen", or "cursor".
-This option is only available on wacom V devices (Intuos and Cintiq series).
-To see which serial number belongs to a device, you need to run the utility program,
+This option is only available on wacom V devices which are capable of reporting a 
+serial number. To see if a connected device is supported, or to identify
+which serial number belongs to a device, you need to run the utility program,
 xsetwacom, that comes with this driver. In layman's terms, this option will add 
 additional X devices for this specific tool's tip (and eraser, if applicable).
 This is useful for programs like gimp (which remembers tools based on the X device) to recall 
--- a/man/wacom.man
+++ b/man/wacom.man
@@ -233,11 +233,11 @@ xsetwacom, which comes with linuxwacom package.
 .B Option \fI"ToolSerials"\fP \fI"number[,type[,label]][;...]"\fP
 sets the list of serial numbered devices that need to be hotplugged for a physical
 device. The 'type' option may be any of "pen", "airbrush", "artpen", or "cursor".
-This option is only available on wacom V devices (Intuos series and Cintiq 21U).
+This option is only available on wacom V devices (Intuos and Cintiq series).
 To see which serial number belongs to a device, you need to run the utility program,
 xsetwacom, that comes with this driver. In layman's terms, this option will add 
 additional X devices for this specific tool's tip (and eraser, if applicable).
-This is useful for programs like gimp (which remembers tools based on X device) to recall 
+This is useful for programs like gimp (which remembers tools based on the X device) to recall 
 additional drawing tool selections for an airbrush+eraser, art pen, extra pen, etc. 
 .TP 4
 .B Option \fI"Threshold"\fP \fI"number"\fP
--- a/man/wacom.man
+++ b/man/wacom.man
@@ -235,7 +235,10 @@ sets the list of serial numbered devices that need to be hotplugged for a physic
 device. The 'type' option may be any of "pen", "airbrush", "artpen", or "cursor".
 This option is only available on wacom V devices (Intuos series and Cintiq 21U).
 To see which serial number belongs to a device, you need to run the utility program,
-xsetwacom, that comes with this driver.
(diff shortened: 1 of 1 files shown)


man/wacom.man
```diff
@@ -235,7 +235,10 @@ sets the list of serial numbered devices that need to be hotplugged for a physic
 device. The 'type' option may be any of "pen", "airbrush", "artpen", or "cursor".
 This option is only available on wacom V devices (Intuos series and Cintiq 21U).
 To see which serial number belongs to a device, you need to run the utility program,
-xsetwacom, that comes with this driver.
+xsetwacom, that comes with this driver. In layman's terms, this option will add X
+(virtual) slave pointers for this specific tool's tip (and eraser, if applicable).
```
whot, line 239: I'd just say "will add an X device for this specific tool", the exact approach how X handles the devices doesn't matter for this purpose.

Plus, we should avoid using the term `slave pointer` as much as possible, even though we'll likely never be able to get rid of it here.

man/wacom.man
```diff
@@ -235,7 +235,10 @@ sets the list of serial numbered devices that need to be hotplugged for a physic
 device. The 'type' option may be any of "pen", "airbrush", "artpen", or "cursor".
 This option is only available on wacom V devices (Intuos series and Cintiq 21U).
```
whot, line 236: heh, I'd say that sentence is a bit ouf of date now...

man/wacom.man
```diff
@@ -235,7 +235,10 @@ sets the list of serial numbered devices that need to be hotplugged for a physic
...
 To see which serial number belongs to a device, you need to run the utility program,
-xsetwacom, that comes with this driver.
+xsetwacom, that comes with this driver. In layman's terms, this option will add X
+(virtual) slave pointers for this specific tool's tip (and eraser, if applicable).
+This is useful for programs like gimp (which remembers tools based on X device) to recall 
+additional drawing tool selections for your airbrush+eraser, art pen, extra pen, etc.
```
whot, line 241: style: we use passive form and avoid "your" in man pages.

man/wacom.man
vintagepc: Such is the nature of documentation ;-)

man/wacom.man
vintagepc: Fixed

man/wacom.man
vintagepc: Would you like me to do something about it? I don't know what additional hardware it applies to...

whot: Yes please. "Intuos Pro and Cintiq series" I think is good enough here, @jigpu?

man/wacom.man
```diff
@@ -235,7 +235,10 @@ sets the list of serial numbered devices that need to be hotplugged for a physic
 device. The 'type' option may be any of "pen", "airbrush", "artpen", or "cursor".
 This option is only available on wacom V devices (Intuos series and Cintiq 21U).
 To see which serial number belongs to a device, you need to run the utility program,
-xsetwacom, that comes with this driver.
+xsetwacom, that comes with this driver. In layman's terms, this option will add 
+additional X devices for this specific tool's tip (and eraser, if applicable).
+This is useful for programs like gimp (which remembers tools based on X device) to recall
```
whot, line 240: just noticed this: grammar: "based on the X device" (missing "the")

vintagepc: I'm not going to argue about it since it is your repo/preferred style, but it's not uncommon to have omitted determiners in clauses like this where there is a [plural] to [plural] relationship with an implied direct 1:1 mapping. For example, it's also common vernacular to say something like "Students' grades were based on [implied "their"] effort"

man/wacom.man
vintagepc: Tweaked, I did not include "Pro" as that's a distinct product line from the Intuos3 I have that supports it.

whot: The intuos3 at this point is 12 years old? And quite a few years ago, Wacom changed the naming, so what used to be Bamboos are now called Intuos and the previous Intuos devices are called Intuos Pro. And the Bamboos don't support serials, so basically: if you leave out the "Pro", it's the wrong device for anything released in the last 8 or so years. And no, don't ask me **why** the naming was changed in the most confusing way possible.

vintagepc: If the feature is only dependent on the device being able to identify tool serials... should we remove that line entirely and replace it with something like "This feature is only supported if xsetwacom can report a tool serial number from the device"?

whot: yeah, that works too. I'm not a big fan of the emphasis on `xsetwacom` though since you can get the same information from e.g. `xinput`. But I guess it's less confusing than writing something more generic...

jigpu: I agree with simplifying the wording, but would want to keep the Cintiq and Intuos Pro examples since most people probably have no clue which devices support tool IDs. Also, the phrase "wacom V devices" is meaningless to anyone who doesn't maintain this driver, and probably even incorrect (I'd bet the feature also works with AES pens that report serial numbers) so I'd remove it as well.

> And no, don't ask me why the naming was changed in the most confusing way possible.

It's a mystery to us all :dizzy_face:

jigpu:
Merging for the next release as this is still an improvement, any remaining nitpicks aside.

PR #137, opened 2020-07-29
Support new Cintiqs for older kernels

Pinglinux:
Kernels older than 2.6.38 do not report INPUT_PROP_DIRECT

Signed-off-by: Ping Cheng <ping.cheng@wacom.com>

### DIFF
--- a/src/wcmValidateDevice.c
+++ b/src/wcmValidateDevice.c
@@ -333,6 +333,10 @@ int wcmDeviceTypeKeys(InputInfoPtr pInfo)
 		case 0x352:/* Cintiq Pro 32 */
 		case 0x37C:/* Cintiq Pro 24 Pen-Only */
 		case 0x390:/* Cintiq 16 */
+		case 0x391:/* Cintiq 22 */
+		case 0x396:/* DTK-1660E */
+		case 0x3AE:/* Cintiq 16 */
+		case 0x3B0:/* DTK-1660E */
 			TabletSetFeature(priv->common, WCM_ROTATION);
 			/* fall-through */
 


src/wcmValidateDevice.c
```diff
@@ -204,9 +204,14 @@ int wcmDeviceTypeKeys(InputInfoPtr pInfo)
...
 		case 0x34E: /* MobileStudio Pro 16 */
+		case 0x390: /* Cintiq 16 */
+		case 0x391: /* Cintiq 22 */
+		case 0x396: /* DTK-1660E */
 		case 0x398: /* MobileStudio Pro 13 */
 		case 0x399: /* MobileStudio Pro 16 */
 		case 0x3AA: /* MobileStudio Pro 16 */
+		case 0x3AE: /* Cintiq 16 */
+		case 0x3B0: /* DTK-1660E */
 			TabletSetFeature(priv->common, WCM_LCD);
```
jigpu, line 215: This doesn't look like the right spot to add these cases. The fall-through will cause these devices to get the `WCM_LCD`, `WCM_ROTATION`, and `WCM_RING` features, but I don't believe these devices have a touchring.

src/wcmValidateDevice.c
```diff
@@ -332,7 +337,6 @@ int wcmDeviceTypeKeys(InputInfoPtr pInfo)
 		case 0x351:/* Cintiq Pro 24 */
 		case 0x352:/* Cintiq Pro 32 */
 		case 0x37C:/* Cintiq Pro 24 Pen-Only */
-		case 0x390:/* Cintiq 16 */
 			TabletSetFeature(priv->common, WCM_ROTATION);
```
jigpu, line 340: I believe this is the right spot for all of these tablets. The fall through here sets only `WCM_LCD` and `WCM_ROTATION`.

Pinglinux: Oh, focused on INPUT_PROP_DIRECT too much and stopped at the first WCM_LCD. Will push the fix.

Issue #139, opened 2020-08-24
If pan scrolling is used with relative mode, the mouse pointer "walks" out of the scrolled region


yump:
When panscroll is active, the cursor moves when the pen is lifted but not when the pen is down.  So when you lift the pen and return to the start position to scroll again, the cursor moves.  This quickly walks the cursor outside of the window you are trying to scroll, so the scroll events stop going to the right place.

To reproduce, configure as follows:

    #!/bin/sh

    stylus="Wacom Intuos S Pen stylus" # replace with your device 

    xinput set-mode "$stylus" RELATIVE
    xsetwacom set "$stylus" Button 2 pan
    xsetwacom set "$stylus" PanscrollThreshold 50

Then hold down the pan button and attempt to scroll several times in the same direction on one window.

The problem is probably more noticeable if your environment uses focus-follows-mouse like mine, but I think web browsers and maybe other applications use mouse position to determine scroll focus for their internal scrollable elements.

Greenscreener:
I have exactly the same problem! This is really annoying.

PR #140, opened 2020-08-24
Stop pointer movement when panscrolling in relative mode

yump:
Always stop pointer movement when a panscroll button is pressed in
relative mode, instead of only when the pen is down. This keeps the
pointer from "walking" when the user lifts the pen and scrolls
repeatedly.

Fixes: https://github.com/linuxwacom/xf86-input-wacom/issues/139
Signed-off-by: Russell Haley <yumpusamongus@gmail.com>

### DIFF
--- a/src/wcmCommon.c
+++ b/src/wcmCommon.c
@@ -640,7 +640,7 @@ wcmSendNonPadEvents(InputInfoPtr pInfo, const WacomDeviceState *ds,
 
 		/* Move the cursor to where it should be before sending button events */
 		if(!(priv->flags & BUTTONS_ONLY_FLAG) &&
-		   !(priv->flags & SCROLLMODE_FLAG && priv->oldState.buttons & 1))
+		   !(priv->flags & SCROLLMODE_FLAG && (!is_absolute(pInfo) || priv->oldState.buttons & 1)))
 		{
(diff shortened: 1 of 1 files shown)


Greenscreener:
Awesome! It could also work by not disabling pointer movement in either case, which is for instance what Krita does with the Pan tool. Could this get merged?

yump:
Not disabling pointer movement would cause problems if you are trying to scroll on a very small UI element, such as a volume control or numeric input box.

In the interests of disclosure, I should not make any changes to this PR because I will not be able to test them, because I have returned the tablet device I borrowed from a friend and decided not to buy one of my own.

jigpu:
Merged.

@Greenscreener did you mean "not disabling" or "disabling"? Not disabling in both modes modes is a bad idea as @yump already outlined, but _disabling_ in both modes (so that the pointer stays frozen while the button is held _even in absolute mode_) could be useful, I think...

Greenscreener:
I meant "not disabling", but I didn't exactly think it through and I fully agree with what @yump said afterwards. "Either case" didn't mean "absolute/relative", it was supposed to mean "pen down/up", many programs do that, where you can see the cursor moving with the content under it, as if it was a physical "hand" dragging it. But still, I agree with @yump, that this would do more bad than good.

PR #162, opened 2021-02-02
Recognize pad devices which only have softkeys (e.g. Cintiq Pro 32)

jigpu:
Devices like the Cintiq Pro 32 have pad devices that have no ExpressKeys.
This prevents them from being recognized as a pad by the driver. When
the device is connected, an `Invalid type 'pad' for this device`
message would be logged.

This has not been an issue in the past since there is nothing for the
device to do without any ExpressKeys. Now that the driver is capable
of forwarding the various softkeys as fixed-function keyboard events,
however, it would be good to ensure that these kinds of devices are
detected properly.

Recognizing the device as a pad only requires us to expand on the list
of tools that are known identifiers. Since we want softkey-only pads to
be recognized, we two softkeys to the list. One or both of the keys
should be present on devices like the Cintiq Pro 32, allowing the pad
to be recognized as such.

### DIFF
--- a/src/wcmValidateDevice.c
+++ b/src/wcmValidateDevice.c
@@ -123,14 +123,14 @@ ret:
 static struct
 {
 	const char* type;
-	__u16 tool[3]; /* tool array is terminated by 0 */
+	__u16 tool[5]; /* tool array is terminated by 0 */
 } wcmType [] =
 {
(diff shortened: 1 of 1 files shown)


Pinglinux (approved):
The patch makes sense to me.

whot (approved):
> we two softkeys to ...

typo in commit message but otherwise LGTM, thanks

PR #190, opened 2021-11-19
Require xserver 1.10, use xf86IDrvMsg

whot:
The first patchset of probably quite a few to come, a few low-hanging fruit.

We bump to xserver 1.10 which is 10y old now and available since RHEL6.2, so safe enough. This gives us `xf86IDrvMsg` which is a more consistent way of logging. The rest is janitorial.

cc @Pinglinux

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,7 @@
 on: [ push, pull_request ]
 
 env:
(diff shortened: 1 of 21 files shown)


Pinglinux:
The patchset looks good to me. This work reminded me of those good old days more than 10 years ago ;). I relied on Peter for all the X server/Xorg/XInput specific functions/API/ABI...

PR #191, opened 2021-11-22
Miscellaneous janitor patches

whot:
Use two standard libc calls instead of their server wrappers and change `wcmOpen()` to have a more sensible behaviour. 

The rest is just tidying up.


Pinglinux:
The set looks good. All patches are:

Thank you Peter!

PR #192, opened 2021-11-23
More cleanup patches

whot:
All over the place, no real functional changes.


Pinglinux:
At some point I should test, instead of just review, the patches. For now, they looked fine.

PR #193, opened 2021-11-23
Rework the action handlers

whot:
Sits on top of #192 , only the last 3 matter

This simplifies the handling of button actions, instead of copying arrays around we just wrap those into a struct that can be passed around easier.

### DIFF
--- a/src/wcmCommon.c
+++ b/src/wcmCommon.c
@@ -353,10 +353,10 @@ static void sendAButton(InputInfoPtr pInfo, const WacomDeviceState* ds, int butt
 	DBG(4, priv, "TPCButton(%s) button=%d state=%d\n",
 	    common->wcmTPCButton ? "on" : "off", button, mask);
 
-	if (wcmActionSize(&priv->keys[button]) == 0)
+	if (wcmActionSize(&priv->key_actions[button]) == 0)
 		return;
 
-	sendAction(pInfo, ds, (mask != 0), &priv->keys[button],
+	sendAction(pInfo, ds, (mask != 0), &priv->key_actions[button],
 		   first_val, num_val, valuators);
 }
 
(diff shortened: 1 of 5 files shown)


whot:
Pushed this on top of #196 now, I'm getting too many branches going at the same time here...

And a bit more cleanup to make all this easier to understand and more logically separated.

Pinglinux:
Since other requests are closed, I'm attaching all my comments here:

1. Should we close this request? The patches have been merged, I think.
2. I tested the updated master branch. I found the some mismatch in the results between the old ./autogen.sh and the new meson/ninja building process.
   2.1 When building from old process, wacom_drv.so is installed to /usr/lib/xorg/modules/input;
   2.2 When building from the new meson process, wacom_drv.so is installed to /usr/local/lib/x86_64-linux-gnu/xorg/modules/input.
3. That is fine as long as the proper driver is loaded. However, when I logout then log back in (or reboot the system), the wacom_drv.so under /usr/lib/xorg/modules/input is always used, no matter which one is the latest built. I got that information from the ~/.loca/share/xorg/Xorg.0.log file. So, the wacom_drv.so under /usr/local/lib/x86_64-linux-gnu/xorg/modules/input is not loaded. How do I test it?
4. Also, xsetwacom is installed under /usr/bin/ by old way. It is under /usr/local/bin by meson. When I which xsetwacom, /usr/local/bin/xsetwacom showed if meson was used. That is fine. However, once meson is used, only the xsetwacom under /usr/local/bin will be used even if I build the driver the old way.
5. So there are mismatch between the two building processes. Are we going to keep both approaches?

whot:
> Since other requests are closed, I'm attaching all my comments here:

fwiw you can still comment on the closed MRs, easier too since it keeps bugs listed in one place.

These patches haven't been merged yet, they're less of just cleanup and need a second pair of eyes to verify nothing's wrong with them.


As for your questions: meson defaults to a prefix of `/usr/local`, `configure` (apparently) to `/usr` in your setup. If you build with the respective prefix, everything should be the same:

```
$  meson configure builddir/ -Dprefix=/usr
$ ninja -C builddir/ install
```

Note: `meson setup` on the first run only, `meson configure` to change an existing builddir's configuration - it's equivalent to autotool's `./configure` run. You can provide `-D` options during `meson setup` as well.

Pinglinux:
Reviewed the first 3 patches. Compiled and installed it on Ubuntu 20.04 with both building processes. Driver works.

PR #194, opened 2021-11-24
Add support for meson

whot:
A bunch of build system cleanups, nothing overly exciting but it paves the way for adding support for meson which is a vastly superior build system to autotools.

Installed output looks the same, did a manual comparison of the `nm` output and except for a few glibc symbols being slighly different the data is the same.

What is explicitly missing is support for HAL fdi files - the X server has had udev for a decade or more now, not sure anyone sill uses HAL (non-linux systems maybe).

Compile and install with:

```
  meson setup builddir/   # optionally: -Dprefix=/whatever and other options
  meson compile -C builddir/
  meson install -C builddir 
```


whot:
Well, that took more changes than expected but it works now

Pinglinux:
> Compile and install with: 
> ```
>   meson setup builddir/   # optionally: -Dprefix=/whatever and other options
>   meson compile -C builddir/
>   meson install -C builddir 
> ```

I tested it on Ubuntu 20.04. The actual commands I used are:

$ sudo apt install meson
$ mkdir mesonbuild
$ meson setup mesonbuild/
$ meson compile ("meson compile -C" led to "meson: error: unrecognized arguments: -C")
$ cd mesonbuild
$ sudo meson install

Looks like the driver built and installed under /usr/local/lib/x86_64-linux-gnu/xorg/modules/input. Are those the exact steps we need to follow? Will you update the wiki page accordingly?

whot:
Nicely enough, meson creates the directory so `mkdir` is superfluous.

`meson <foo>` will work if you `cd`  into the directory first, otherwise `-C` is required. Not sure why this didn't work on your box, I suspect you ended up with this sequence (`cd` before `meson foo`):
```
$ meson setup mesonbuild
$ cd mesonbuild
$ meson compile
$ sudo meson install
```
The above works, because `-C` is simply "change to this directory first". Maybe ubuntu has an older version where this doesn't work? I don't know when that `compile` command was added to meson but I'm pretty sure it was before whatever version we require.

And yes, I'll update the wiki eventually, but let's let this settle first before we push users to use it - just in case some bugs sneaked in.

whot:
Ok, tested this and indeed, that version (0.53.0) doesn't support `-C` for the `meson compile` command. I'll amend the commit message to use `ninja` directly so we don't have to worry about different versions here.

whot:
Merging this, thanks for the testing!

PR #195, opened 2021-11-24
Buildsystem cleanups

whot:
This is the cleanup part in preparation for meson support (#194),  just so we can merge one before the other if needed


whot:
Merged as part of #194

PR #196, opened 2021-11-24
Require xserver 1.13

whot:
This sits on top of #194  and #195.

1.13 is 9 years old by now and it's available in RHEL6.4 and later, so we should be good here. Most of the ABI changes happened before that release, so we can drop most of our ifdefs.


whot:
Merging this one, it's harmless enough unless you're still on RHEL5 in which case... well.. sorry.

PR #198, opened 2021-11-29
General cleanup with a few odd fixes

whot:
This sits on top of #193 but overall it's just cleanup with no functional changes to the driver.

Notable changes:
- we're down to very few ifdefs now after requiring kernel 2.6.30 and #defining the remaining few
- we now store the input props in the driver too - we only ever check one of those but the few bytes needed won't hurt us

The rest is just cleanup, some functions moved and/or slightly changed to make them easier to follow, some bits no longer needed were dropped, etc.


Pinglinux:
There are conflicts in src/wcmCommon.c...

whot:
> There are conflicts in src/wcmCommon.c...

weird. not sure why, git diff didn't show anything. maybe just github being stupid.

PR #201, opened 2021-12-01
Enable subdir-objects to stop autotools from complaining

whot:
test/../src/common.mk:3: warning: source file '$(top_srcdir)/src/wcmCommon.c' is in a subdirectory,
test/../src/common.mk:3: but option 'subdir-objects' is disabled
test/Makefile.am:2:   'test/../src/common.mk' included from here

etc


Pinglinux:
It's nice to get rid of those annoying messages. Thank you, Peter!

PR #203, opened 2021-12-02
release script updates

whot:
A few minor cleanups and the addition of support for meson which we'll need after https://github.com/linuxwacom/libwacom/pull/445

There's likely to be a bug we'll find once we run it, like usual, but dry-runs only get you so far :)

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -143,8 +143,8 @@ release_to_github() {
     # Upload the tar to the release
     upload_result=`curl -s -u $GH_USERNAME \
         -H "Content-Type: application/x-bzip" \
-        --data-binary @$tarbz2 \
-        "https://uploads.github.com/repos/$GH_REPO/$PROJECT/releases/$GH_RELEASE_ID/assets?name=$tarbz2"`
+        --data-binary @$tarball \
+        "https://uploads.github.com/repos/$GH_REPO/$PROJECT/releases/$GH_RELEASE_ID/assets?name=$tarball"`
     DL_URL=`echo $upload_result | jq -r '.browser_download_url'`
 
     check_json_message "$upload_result"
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -434,43 +432,57 @@ process_module() {
...
     # Important to run make dist/distcheck before looking in Makefile, may need to reconfigure
-    echo "Info: running \"make $MAKE_DIST_CMD\" to create tarballs:"
-    ${MAKE} $MAKEFLAGS $MAKE_DIST_CMD > /dev/null
-    if [ $? -ne 0 ]; then
-	echo "Error: \"$MAKE $MAKEFLAGS $MAKE_DIST_CMD\" failed."
-	cd $top_src
-	return 1
+    if $use_meson; then
+	echo "Info: running \"meson dist\" to create tarballs:"
+	ninja dist
```
jigpu, line 437: Presumably the status message should say "ninja" here?

release.sh
```diff
@@ -434,43 +432,57 @@ process_module() {
 
     # Run 'make dist/distcheck' to ensure the tarball matches the git module content
     # Important to run make dist/distcheck before looking in Makefile, may need to reconfigure
-    echo "Info: running \"make $MAKE_DIST_CMD\" to create tarballs:"
-    ${MAKE} $MAKEFLAGS $MAKE_DIST_CMD > /dev/null
-    if [ $? -ne 0 ]; then
-	echo "Error: \"$MAKE $MAKEFLAGS $MAKE_DIST_CMD\" failed."
-	cd $top_src
-	return 1
+    if $use_meson; then
```
jigpu, line 435: Can we indent the if/else blocks here? It almost caused me to miss the else block entirely...

whot: they're indented correctly in the source file but unfortunately we're mixing 4-space indentation with tabs for 8 spaces, and github in its limited wisdom renders tabs at 4 spaces. I'll send a follow-up to fix indentation to spaces only in this file once this is merged to avoid this in the future.

jigpu (approved):

whot:
merged, thanks for the review!

PR #211, opened 2022-01-04
Remove support for serial devices from the driver

### LINKED ISSUE #197: Keep the serial code alive?
I was reworking the driver a fair bit recently and I'm now wondering: is it worth keeping the ISDV4 serial backend alive? When was the last time Wacom released a compatible device for this code?

cc @Pinglinux, @jigpu

whot:
The serial code inside the driver is effectively unused and definitely
untested. It's been a decade since the last serial devices came out and few of
those will be in active use, the few that are are likely supported by
the inputattach code we keep in place for now.

Fixes #197


jigpu:
Works for me :)

PR #213, opened 2022-01-07
Report puck "throttle" (absolute scrollwheel) data in correct axis

jigpu:
The "throttle" / absolute scrollwheel found on devices like the 4D Mouse
should be reported in the v4 axis. A typo in a previous commit caused
it to stop being reported.

Fixes: 58a931bc21 ("Abstract the event interface to pass a struct with axis data around")


whot (approved):
whoopsie, sorry

PR #218, opened 2022-01-24
meson: default to the module dir from xorg-server.pc

### LINKED ISSUE #217: "sudo meson install" copied wacom_drv.so to a wrong path
Cloned the latest xf86-input-wacom. Then followed the exact steps for "Building with meson". wacom_drv.so is copied to /usr/lib/x86_64-linux-gnu/xorg/modules/input, instead of the existing /usr/lib/xorg/modules/input.
I guess we either need to update the build instruction or the configure script.

whot:
On Ubuntu 20.04 at least, libdir is /usr/lib/x86_64-linux-gnu/ but the
xorg modules are in /usr/lib/xorg/modules/. A default install with
prefix /usr thus puts the modules into the wrong directory.

Fix this by defaulting to the xorg module path from the xserver
pkgconfig file. That has been exported since 2005, so it's save to rely
on.

Fixes #217

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -9,7 +9,6 @@ wacom_version = meson.project_version().split('.')
 dir_man1	= get_option('prefix') / get_option('mandir') / 'man1'
 dir_man4	= get_option('prefix') / get_option('mandir') / 'man4'
 dir_bin 	= get_option('prefix') / get_option('bindir')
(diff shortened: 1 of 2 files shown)


Pinglinux:
Tested and works. I also built the driver using the existing "autotools" commands. That approach still works ;).

PR #222, opened 2022-02-01
Implement smooth panscrolling

### LINKED ISSUE #220: Synaptics-like smoothness for PanScrolling
I have a One by Wacom M, that I use the PanScrolling feature on. (`xsetwacom set "..." Button 2 pan`).
This feature is great and allows me to navigate the UI of almost all programs very naturally. 
There's however one thing, that annoys me slightly: For panscrolling to be usable, you have to set the PanScrollThreshold really high (it's 150 for me) which means the pen moves 150 dots, sends a single large scroll event, another 150 dots, another large scroll event, which makes scrolling really choppy. When I scroll on my touchpad, scrolling is much smoother and doesn't happen in (noticeable) discrete steps. 

The behaviour demonstrated in xournalpp:

https://user-images.githubusercontent.com/22398555/151708200-518ad2f4-b81d-42e6-96a4-7e2c576c3931.mp4

If velocity could be implemented as well, that would be a very nice touch, but that isn't that useful for tablets as it is useful for touchpads. 

After doing some research, this is apparently done using valuators 2 and 3 in an X motion event. I will attempt to implement this myself and post a PR, but I'm not very experienced, so we'll see how that goes.

Greenscreener:
Hello, 
with @whot's [advice](https://github.com/linuxwacom/xf86-input-wacom/issues/220#issuecomment-1025340648) I created a basic implementation of smooth scrolling for panscrolling. It is very much a first draft. 
TODO:
 - [x] fix bugs
   - [x] scrolling is "springy" it snaps back into the original place after letting go in absolute mode
   - [x] ~~some modification I made breaks relative mode entirely - the data passed to the valuators seems to be absolute, which makes the cursor snap to the bottom right  corner of the screen~~ #226 
   - [x] ~~xournalpp~~ gtk apps register only the beginning of a touch for some reason, only makes dots and doesn't make proper strokes, hover events are weird
   - [x] ~~qt apps don't register the scrolling at all~~ #247 
 - [x] make `PanScrollThreshold` the option to set the scroll speed and make them cover the same distance at the same value
 - [x] ~~man+docs~~ since no options were modified, there's nothing to add
 - [x] ~~tests~~
 - [x] remove `DBG(0...`
 - [x] history rewrite - split into three commits
   - [x] fixing nbaxes discrepancy
   - [x] moving from valuator array to valuator mask in convertAxes
   - [x] implementing smooth panscrolling

 If anyone could give me some more bits of advice on how to tackle these issues, that would be great ; @whot's three paragraphs helped me immensely. 

Thank you for your help and all the work you've put into this driver!
GS

Fixes #220

### DIFF
--- a/src/WacomInterface.h
+++ b/src/WacomInterface.h
@@ -63,8 +63,10 @@ enum WacomAxisType {
 	WACOM_AXIS_WHEEL	= (1 << 9), /* Artpen rotation or airbrush wheel */
 	WACOM_AXIS_RING		= (1 << 10),
 	WACOM_AXIS_RING2	= (1 << 11),
+	WACOM_AXIS_SCROLL_X	= (1 << 12),
+	WACOM_AXIS_SCROLL_Y	= (1 << 13),
 
-	_WACOM_AXIS_LAST = WACOM_AXIS_RING2,
+	_WACOM_AXIS_LAST = WACOM_AXIS_SCROLL_Y,
 };
 
 typedef struct {
@@ -77,6 +79,7 @@ typedef struct {
 	int throttle;
 	int wheel;
 	int ring, ring2;
+	int scroll_x, scroll_y;
 } WacomAxisData;
 
 
@@ -163,6 +166,8 @@ static inline void wcmAxisSet(WacomAxisData *data,
 	case WACOM_AXIS_WHEEL: data->wheel = value; break;
 	case WACOM_AXIS_RING: data->ring = value; break;
 	case WACOM_AXIS_RING2: data->ring2 = value; break;
+	case WACOM_AXIS_SCROLL_X: data->scroll_x = value; break;
+	case WACOM_AXIS_SCROLL_Y: data->scroll_y = value; break;
 	default:
 		abort();
 	}
@@ -187,6 +192,8 @@ static inline bool wcmAxisGet(const WacomAxisData *data,
 	case WACOM_AXIS_WHEEL: *value_out = data->wheel; break;
 	case WACOM_AXIS_RING: *value_out = data->ring; break;
 	case WACOM_AXIS_RING2: *value_out = data->ring2; break;
+	case WACOM_AXIS_SCROLL_X: *value_out = data->scroll_x; break;
+	case WACOM_AXIS_SCROLL_Y: *value_out = data->scroll_y; break;
 	default:
 		abort();
 	}
@@ -224,6 +231,8 @@ static inline const char* wcmAxisName(enum WacomAxisType which)
 	case WACOM_AXIS_WHEEL: return "wheel";
 	case WACOM_AXIS_RING: return "ring";
 	case WACOM_AXIS_RING2: return "ring2";
+	case WACOM_AXIS_SCROLL_X: return "scroll-x";
+	case WACOM_AXIS_SCROLL_Y: return "scroll-y";
 	default:
 		abort();
 	}
--- a/src/wcmConfig.c
+++ b/src/wcmConfig.c
@@ -98,6 +98,9 @@ WacomDevicePtr wcmAllocate(void *frontend, const char *name)
 	priv->tap_timer = wcmTimerNew();
 	priv->touch_timer = wcmTimerNew();
 
+	/* reusable valuator mask */
+	priv->valuator_mask = valuator_mask_new(7);
+
 	return priv;
 
 error:
--- a/src/wcmConfig.c
+++ b/src/wcmConfig.c
@@ -99,7 +99,7 @@ WacomDevicePtr wcmAllocate(void *frontend, const char *name)
 	priv->touch_timer = wcmTimerNew();
 
 	/* reusable valuator mask */
-	priv->valuator_mask = valuator_mask_new(7);
+	priv->valuator_mask = valuator_mask_new(8);
 
 	return priv;
 
@@ -1220,6 +1220,14 @@ static int wcmInitAxes(WacomDevicePtr priv)
 		wcmInitAxis(priv, WACOM_AXIS_RING2, min, max, res);
 	}
 
+	if (IsPen(priv)) {
+		/* seventh valuator: scroll_x */
+		wcmInitAxis(priv, WACOM_AXIS_SCROLL_X, -1, -1, 0);
+
+		/* eighth valuator: scroll_y */
+		wcmInitAxis(priv, WACOM_AXIS_SCROLL_Y, -1, -1, 0);
+	}
+
 	return TRUE;
 }
 
@@ -1232,13 +1240,16 @@ Bool wcmDevInit(WacomDevicePtr priv)
 	if (priv->common->wcmModel->DetectConfig)
 		priv->common->wcmModel->DetectConfig (priv);
 
-	nbaxes = priv->naxes;       /* X, Y, Pressure, Tilt-X, Tilt-Y, Wheel */
+	nbaxes = priv->naxes;       /* X, Y, Pressure, Tilt-X, Tilt-Y, Wheel, Scroll-X, Scroll-Y */
 	if (!nbaxes || nbaxes > 6)
 		nbaxes = priv->naxes = 6;
 	nbbuttons = priv->nbuttons; /* Use actual number of buttons, if possible */
 
 	if (IsPad(priv) && TabletHasFeature(priv->common, WCM_DUALRING))
 		nbaxes = priv->naxes = nbaxes + 1; /* ABS wheel 2 */
+
+	if (IsPen(priv))
+		nbaxes = priv->naxes = nbaxes + 2; /* Scroll X and Y */
 
 	/* if more than 3 buttons, offset by the four scroll buttons,
 	 * otherwise, alloc 7 buttons for scroll wheel. */
--- a/src/x11/xf86Wacom.c
+++ b/src/x11/xf86Wacom.c
@@ -388,6 +388,8 @@ valuatorNumber(enum WacomAxisType which)
 	case WACOM_AXIS_WHEEL: pos = 5; break;
 	case WACOM_AXIS_RING: pos = 5; break;
 	case WACOM_AXIS_RING2: pos = 6; break;
+	case WACOM_AXIS_SCROLL_X: pos = 6; break;
+	case WACOM_AXIS_SCROLL_Y: pos = 7; break;
 		break;
 	default:
 		abort();
--- a/src/x11/xf86Wacom.c
+++ b/src/x11/xf86Wacom.c
@@ -518,15 +518,28 @@ void wcmInitAxis(WacomDevicePtr priv, enum WacomAxisType type,
 			break;
 		case WACOM_AXIS_RING2:
 			break;
+		case WACOM_AXIS_SCROLL_X:
+			label = XIGetKnownProperty(AXIS_LABEL_PROP_REL_HSCROLL);
+			break;
+		case WACOM_AXIS_SCROLL_Y:
+			label = XIGetKnownProperty(AXIS_LABEL_PROP_REL_VSCROLL);
+			break;
+
 		default:
 			abort();
 	}
 
 	index = valuatorNumber(type);
 	InitValuatorAxisStruct(pInfo->dev, index,
-	                       label,
-	                       min, max, res, min_res, max_res,
-	                       Absolute);
+			       label,
+			       min, max, res, min_res, max_res,
+			       Absolute);
+
+	if (type == WACOM_AXIS_SCROLL_X)
+		SetScrollValuator(pInfo->dev, index, SCROLL_TYPE_HORIZONTAL, PANSCROLL_INCREMENT, 0);
+	else if (type == WACOM_AXIS_SCROLL_Y)
+		SetScrollValuator(pInfo->dev, index, SCROLL_TYPE_VERTICAL, PANSCROLL_INCREMENT, 0);
+
 }
 
 bool wcmInitButtons(WacomDevicePtr priv, unsigned int nbuttons)
@@ -1119,7 +1132,7 @@ valuator_mask_get(const ValuatorMask *mask, int valuator)
 TEST_CASE(test_convert_axes)
 {
 	WacomAxisData axes = {0};
-	ValuatorMask *mask = valuator_mask_new(7);
+	ValuatorMask *mask = valuator_mask_new(8);
 
 	convertAxes(&axes, mask);
 	assert(valuator_mask_num_valuators(mask) == 0);
@@ -1165,6 +1178,7 @@ TEST_CASE(test_convert_axes)
 	assert(valuator_mask_isset(mask, 5));
 	assert(valuator_mask_get(mask, 5) == 2);
 	assert(!valuator_mask_isset(mask, 6));
+	assert(!valuator_mask_isset(mask, 7));
 
 	memset(&axes, 0, sizeof(axes));
 	valuator_mask_zero(mask);
@@ -1195,6 +1209,7 @@ TEST_CASE(test_convert_axes)
 	assert(valuator_mask_get(mask, 5) == 2);
 	assert(!valuator_mask_isset(mask, 6));
 	assert(!valuator_mask_isset(mask, 7));
+	assert(!valuator_mask_isset(mask, 8));
 
 	free(mask);
 }
--- a/src/x11/xf86Wacom.c
+++ b/src/x11/xf86Wacom.c
@@ -397,11 +397,9 @@ valuatorNumber(enum WacomAxisType which)
 }
 
 static inline void
-convertAxes(const WacomAxisData *axes, int *first_out, int *num_out, int valuators_out[7])
+convertAxes(const WacomAxisData *axes, ValuatorMask *mask)
 {
-	int first = 7;
-	int last = -1;
-	int valuators[7] = {0};
+
 
 	for (enum WacomAxisType which = _WACOM_AXIS_LAST; which > 0; which >>= 1)
 	{
@@ -414,51 +412,43 @@ convertAxes(const WacomAxisData *axes, int *first_out, int *num_out, int valuato
 
 		/* Positions need to match wcmInitAxis */
 		pos = valuatorNumber(which);
-		first = min(first, pos);
-		last = max(last, pos);
-		valuators[pos] = value;
+		valuator_mask_set(mask, pos, value);
 	}
-
-	if (last < 0)
-		first = 0;
-	*first_out = first;
-	*num_out = last - first + 1;
-
-	memcpy(valuators_out, &valuators[first], *num_out * sizeof(valuators[0]));
 }
 
 void wcmEmitProximity(WacomDevicePtr priv, bool is_proximity_in,
 		      const WacomAxisData *axes)
 {
 	InputInfoPtr pInfo = priv->frontend;
-	int valuators[7] = {0};
-	int first_val, num_vals;
 
-	convertAxes(axes, &first_val, &num_vals, valuators);
+	ValuatorMask *mask = priv->valuator_mask;
+	valuator_mask_zero(mask);
+	convertAxes(axes, mask);
 
-	xf86PostProximityEventP(pInfo->dev, is_proximity_in, first_val, num_vals, valuators);
+	xf86PostProximityEventM(pInfo->dev, is_proximity_in, mask);
 }
 
 void wcmEmitMotion(WacomDevicePtr priv, bool is_absolute, const WacomAxisData *axes)
 {
 	InputInfoPtr pInfo = priv->frontend;
-	int valuators[7] = {0};
-	int first_val, num_vals;
 
-	convertAxes(axes, &first_val, &num_vals, valuators);
+	ValuatorMask *mask = priv->valuator_mask;
+	valuator_mask_zero(mask);
+	convertAxes(axes, mask);
 
-	xf86PostMotionEventP(pInfo->dev, is_absolute, first_val, num_vals, valuators);
+	xf86PostMotionEventM(pInfo->dev, is_absolute, mask);
 }
 
 void wcmEmitButton(WacomDevicePtr priv, bool is_absolute, int button, bool is_press, const WacomAxisData *axes)
 {
 	InputInfoPtr pInfo = priv->frontend;
-	int valuators[7] = {0};
-	int first_val, num_vals;
 
-	convertAxes(axes, &first_val, &num_vals, valuators);
+	ValuatorMask *mask = priv->valuator_mask;
+	valuator_mask_zero(mask);
+	convertAxes(axes, mask);
 
-	xf86PostButtonEventP(pInfo->dev, is_absolute, button, is_press, first_val, num_vals, valuators);
+
+	xf86PostButtonEventM(pInfo->dev, is_absolute, button, is_press, mask);
 }
 
 void wcmEmitTouch(WacomDevicePtr priv, int type, unsigned int touchid, int x, int y)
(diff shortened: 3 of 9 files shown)


src/WacomInterface.h
```diff
@@ -63,8 +63,10 @@ enum WacomAxisType {
 	WACOM_AXIS_WHEEL	= (1 << 9),
 	WACOM_AXIS_RING		= (1 << 10),
 	WACOM_AXIS_RING2	= (1 << 11),
+	WACOM_AXIS_SCROLL_X	= (1 << 12),
+	WACOM_AXIS_SCROLL_Y	= (1 << 13),
```
whot, line 67: fwiw, I found using "vertical" and "horizontal" (or just v an h) to be less likely to introduce typos when it comes to scrolling. That' just a general handwavy comment though.

src/x11/xf86Wacom.c
```diff
@@ -401,9 +401,11 @@ convertAxes(const WacomAxisData *axes, int *first_out, int *num_out, int valuato
...
 		case WACOM_AXIS_ROTATION: pos = 3; break;
 		case WACOM_AXIS_THROTTLE: pos = 4; break;
-		case WACOM_AXIS_WHEEL: pos = 5; break;
-		case WACOM_AXIS_RING: pos = 5; break;
-		case WACOM_AXIS_RING2: pos = 6; break;
+		case WACOM_AXIS_SCROLL_X: pos = 5; break;
+		case WACOM_AXIS_SCROLL_Y: pos = 6; break;
+		case WACOM_AXIS_WHEEL: pos = 7; break;
+		case WACOM_AXIS_RING: pos = 8; break;
+		case WACOM_AXIS_RING2: pos = 8; break;
```
whot, line 408: I've only skimmed this but: those axis positions are out of sync with the comment (e.g. `ninth valuator` is scroll y but here it's the fifth).

And given the wacom driver's age and somewhat quirky behaviours it's safest not to touch the existing valuators and just append yours.

**edit:** ok, this is fixed in the follow-up commit but my comment above about not touching the existing ones applies.

src/x11/xf86Wacom.c
```diff
@@ -525,22 +531,39 @@ void wcmInitAxis(WacomDevicePtr priv, enum WacomAxisType type,
...
-	                       label,
-	                       min, max, res, min_res, max_res,
-	                       Absolute);
+
+	if (type == WACOM_AXIS_SCROLL_X) {
+		DBG(0, priv, "Initting x scroll axis: (pInfo->dev, %d, %x, 120, 0)\n", index, SCROLL_TYPE_HORIZONTAL);
+		SetScrollValuator(pInfo->dev, index, SCROLL_TYPE_HORIZONTAL, 120, 0);
+	} else if (type == WACOM_AXIS_SCROLL_Y) {
+		DBG(0, priv, "Initting y scroll axis: (pInfo->dev, %d, %x, 120, 0)\n", index, SCROLL_TYPE_VERTICAL);
+		SetScrollValuator(pInfo->dev, index, SCROLL_TYPE_VERTICAL, 120, 0);
```
whot, line 554: note: `SetScrollValuator` is **in addition** to `InitValuatorAxisStruct`. It's like an extra flag of "use this valuator for scrolling"

src/x11/xf86Wacom.c
```diff
@@ -441,12 +441,7 @@ void wcmEmitMotion(WacomDevicePtr priv, bool is_absolute, const WacomAxisData *a
 	int first_val, num_vals;
 
 	convertAxes(axes, &first_val, &num_vals, valuators);
-	DBG(0, priv, "Emitting motion: [%d,%d,%d,%d,%d,%d,%d,%d,%d]\n",valuators[0],valuators[1],valuators[2],valuators[3],valuators[4],valuators[5],valuators[6],valuators[7],valuators[8]);
-	int x,y;
-	wcmAxisGet(axes, WACOM_AXIS_SCROLL_X, &x);
-	wcmAxisGet(axes, WACOM_AXIS_SCROLL_Y, &y);
-	DBG(0, priv, "x=%d, y=%d\n", x, y);
-	xf86PostMotionEventP(pInfo->dev, is_absolute, first_val, num_vals, valuators);
+	xf86PostMotionEventP(pInfo->dev, is_absolute, first_val, num_vals, &valuators[0]+first_val); // I think this was supposed to be here in the first place, because without this the function doesn't work for first_val != 0.
```
whot, line 444: yes, this looks like a bug in the original code not triggered yet, I'll get this fixed. good find!

src/x11/xf86Wacom.c
```diff
@@ -435,18 +437,22 @@ void wcmEmitProximity(WacomDevicePtr priv, bool is_proximity_in,
 void wcmEmitMotion(WacomDevicePtr priv, bool is_absolute, const WacomAxisData *axes)
 {
 	InputInfoPtr pInfo = priv->frontend;
-	int valuators[7];
+	int valuators[9];
```
whot, line 440: in hindsight, we should really have a `#define` for that number...

whot:
>  the data passed to the valuators seems to be absolute, which makes the cursor snap to the bottom right corner of the screen

`xf86PostMotionEventP` takes a `is_absolute` boolean, you pass in `False` and this should work. Setting up the valuator as relative axis (min == max == -1) is a good idea too then. The springiness my be related to this too btw.

>  implement a configuration option for scroll speed

shouldn't be needed if you keep the existing pan scrolling option. You scale into that range and that controls your scroll speed. Because you can't change the increment after `SetScrollValuator` I recommend picking some high enough constant or that value (e.g. 0xffff) and then scaling into that range based on the pan scroll option. IOW, pan scroll 1000, movement of 50 -> `50/1000 * 0xffff` and that's the value you emit as motion delta.

> implement a configuration option to switch between old and smooth scrolling 

we don't need that one. Once the new method works, the old events will be emulated by the X server for us - the only need to switch to the old version is a bug (can be fixed) or some setup where the pan scrolling was misused for something else (e.g. pan scrolling mapped to button X which is mapped to an action other than scrolling).

src/x11/xf86Wacom.c
Greenscreener: Is it? I based my code on this: https://github.com/freedesktop/xorg-xf86-input-libinput/blob/62f267a952208005bf378c554fa84f4dc2489b35/src/xf86libinput.c#L938, which doesn't call `InitValuatorAxisStruct` on the scroll ones. Is that an error on their side that is silently ignored? Is there any documentation or reference on this?

src/x11/xf86Wacom.c
Greenscreener: I did that because the last valuator (abswheel2) only gets initialized if it is available and thus has to be last, if we want to keep statically defined valuator numbers and not define them at runtime, abswheel2 has to be the last one.

src/x11/xf86Wacom.c
Greenscreener: Is it? I based my code on this: https://github.com/freedesktop/xorg-xf86-input-libinput/blob/62f267a952208005bf378c554fa84f4dc2489b35/src/xf86libinput.c#L938, which doesn't call `InitValuatorAxisStruct` on the scroll ones. Is that an error on their side that is silently ignored? Is there any documentation or reference on this?

Greenscreener:
>  some modification I made breaks relative mode entirely - the data passed to the valuators seems to be absolute, which makes the cursor snap to the bottom right corner of the screen

This actually isn't related to any of my changes. Creating a separate issue: #226

src/x11/xf86Wacom.c
whot: hmm, that is actually a bug in that driver. it just happened to not be a problem because the defaults used for valuators happen to match what we need and the valuator semi-exists because we already initialized the number of valuators earlier.

Either way, it's not technically correct, just happened to work.

src/x11/xf86Wacom.c
whot: Fixed in #224

Greenscreener:
> scrolling is "springy" it snaps back into the original place after letting go in absolute mode

I think this is caused by the fact that when the rel/abs mode switches, all the valuators get set to that mode, the scroll valuators should be always relative, but in absolute mode, they get set to absolute.

Greenscreener:
This is apparently caused by calling `SetScrollValuator`? When the call is removed, the problem disappears. Weird.

Greenscreener:
Ok! The problem appears when scroll valuators are sent in the same motionEvent as other valuators. If the x,y,pressure event doesn't contain the scroll valuators, it works correctly, if the event contains x,y,pressure,...,scrollx,scrolly,... (even if they're zero) it creates the above mentioned bug. This means that the scroll valuators have to be last, but if I [understand correctly](https://github.com/linuxwacom/xf86-input-wacom/pull/222#discussion_r797407167), abswheel2 also has to be last, which is clearly a requirement that cannot be satisfied.

Greenscreener:
I found a possible solution to this, but I have no idea if this won't break something else. I naïvely think it shouldn't.

Greenscreener:
This also fixes absolute mode springiness.

Greenscreener:
>  qt apps don't register the scrolling at all

With `QT_LOGGING_RULES='qt.qpa.input.*=true'`, this is what gets written out when the program (dolphin in this case) starts, seems correct:
```
qt.qpa.input.devices: input device  Wacom One by Wacom M Pen stylus ID 9
qt.qpa.input.devices:    has 8 buttons
qt.qpa.input.devices:    it's a keyboard
qt.qpa.input.devices:    has valuator "Abs X" recognized? true
qt.qpa.input.devices:    has valuator "Abs Y" recognized? true
qt.qpa.input.devices:    has valuator "Abs Pressure" recognized? true
qt.qpa.input.devices:    has valuator "Abs Tilt X" recognized? true
qt.qpa.input.devices:    has valuator "Abs Tilt Y" recognized? true
qt.qpa.input.devices:    has valuator "Rel Horiz Scroll" recognized? true
qt.qpa.input.devices:    has valuator "Rel Vert Scroll" recognized? true
qt.qpa.input.devices:    has valuator "Abs Wheel" recognized? true
qt.qpa.input.devices:    it's a tablet with pointer type "pen"
qt.qpa.input.devices:    it's a scrolling device
```

This gets logged while moving the cursor around:
```
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1440 detail 0 time 175183453 pos  793.4,  511.7 root pos  793.4,  564.7 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1440 detail 0 time 175183461 pos  788.2,  511.8 root pos  788.2,  564.8 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1440 detail 0 time 175183469 pos  783.1,  511.1 root pos  783.1,  564.1 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1440 detail 0 time 175183475 pos  779.0,  509.2 root pos  779.0,  562.2 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1440 detail 0 time 175183482 pos  776.3,  506.1 root pos  776.3,  559.1 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1440 detail 0 time 175183491 pos  775.2,  502.0 root pos  775.2,  555.0 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
```

This gets logged while attempting to scroll:
```
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 7 time 175240884 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 7 time 175240884 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 0 time 175240890 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 0 time 175240899 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 0 time 175240907 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 5 time 175240907 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 5 time 175240907 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 0 time 175240915 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 0 time 175240920 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 0 time 175240929 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 6 time 175240929 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
qt.qpa.input.events: XI2 event on tablet 9 with tool Stylus type Pen seq 1444 detail 6 time 175240929 pos  853.3,  459.5 root pos  853.3,  512.5 buttons 0x0 pressure 0.00 tilt 0, 0 rotation   0.00 modifiers 0x0
```
Position stays the same, the only thing that updates is the `detail` field which seems to be manifesting the backwards-compatibility scroll button events sent by X. 

For comparison this is what moving the cursor does on a touchpad:
```
qt.qpa.input.events: XI2 mouse motion 953,474, time 175362557, source MouseEventNotSynthesized
qt.qpa.input.events: XI2 mouse motion 951,464, time 175362620, source MouseEventNotSynthesized
qt.qpa.input.events: XI2 mouse motion 949,454, time 175362632, source MouseEventNotSynthesized
```
and scrolling:
```
qt.qpa.input.events: scroll wheel @ window pos QPoint(820,535) delta px QPoint(43,58) angle QPoint(43,58)
qt.qpa.input.events: XI2 mouse motion 820,535, time 175396448, source MouseEventNotSynthesized
qt.qpa.input.events: scroll wheel @ window pos QPoint(820,535) delta px QPoint(52,9) angle QPoint(52,9)
qt.qpa.input.events: XI2 mouse motion 820,535, time 175396473, source MouseEventNotSynthesized
qt.qpa.input.events: scroll wheel @ window pos QPoint(820,535) delta px QPoint(71,-48) angle QPoint(71,-48)
qt.qpa.input.events: XI2 mouse motion 820,535, time 175396498, source MouseEventNotSynthesized
qt.qpa.input.events: scroll wheel @ window pos QPoint(820,535) delta px QPoint(53,-57) angle QPoint(53,-57)
qt.qpa.input.events: XI2 mouse motion 820,535, time 175396523, source MouseEventNotSynthesized
qt.qpa.input.events: scroll wheel @ window pos QPoint(820,535) delta px QPoint(12,-43) angle QPoint(12,-43)
```

whot:
Sorry, still ETIME to look at this (probably until *after* the 1.0 release) but one of the things you should *really* look into is updating the gwacom bits and `wacom-record` so you can debug this easier. It's a lot simpler to just run that tool and look at the yaml output than having to deal with qt or xinput or whatever.

Second - record a sequence with `libinput record` and replay it with `libinput replay`, see [here](https://wayland.freedesktop.org/libinput/doc/latest/tools.html). This way you always get the same event sequence at the kernel level which obv. makes debugging a lot easier as there's no variation you need to worry about. The [hid-recorder](https://gitlab.freedesktop.org/libevdev/hid-tools/) and `hid-replay` tools would work too, they're a level lower and exercise the kernel driver too.

Finally, you may wan to look at #241 because that's where you could add tests to verify specific behaviours.

Greenscreener:
Sorry for a longer delay on this, I actually found out that, weirdly, the qt scrolling issue is actually consistent even with the currently distributed version (for me it is 0.40.0). That is particularly odd, while the original version of the driver does scrolling entirely differently. I'll open a separate issue for this. I'll also look into it more tomorrow.

Greenscreener:
Started working on pytest in https://github.com/Greenscreener/xf86-input-wacom/tree/panscroll%2Bpytest

Greenscreener:
@whot could this please get merged with the tests being implemented separately after #241 is done? I feel like the longer this is open, the more merge conflicts and bugs get introduced.

jigpu:
Made some comments on the tests. Hope you are able to find some time to return to this soon, @Greenscreener :)

Greenscreener:
Yeah I do definitely plan on getting back to this, there are sadly some IRL things that have taken priority rn and I want to give this my full concentration when I get back to it. It might take a few weeks though, can't really say at the moment.

Greenscreener:
Finally, I am able to spend some time on this. I cherry-picked the test code from panscroll+pytest, since the pytest stuff has now been merged to master. 

I was able to fix the issue outlined in the cherry-picked commit. Turns out that `sudo` strips the `LD_LIBRARY_PATH` env variable even with `-E`, so it has to be passed directly to the command. I created a little script to run pytest easier in the root of the repo, just for my own usage and testing right now. 

I have however, run into an issue I haven't been able to fix. It seems to me, that the options defined at the beginning of the function in `opts` don't actually get applied to the virtual device that gets created. As you can see, I inserted an interactive shell in my test to be able to play around with the virtual device. If I run `xsetwacom -s get 19 all` (19 is the id of the device in my system) the two options - `Button2` and `PanScrollThreshold` seem to be set to their default values. (`button +2 ` and  `2600` respecitvely) Also, if I use IPython to run `monitor.wacom_device.get_options().get("PanScrollThreshold")` it is correctly set to `150`, but all the other default options are not present - for instance `monitor.wacom_device.get_options().get("Threshold")` returns nothing. 

This is also confirmed by the fact, that when I run `xinput test-xi2 --root 19`, the RawMotion events look like normal cursor movement events and not scroll events.

whot:
>  created a little script to run pytest easier in the root of the repo, just for my own usage and testing right now.

fwiw, master now has a script too which you can run with `sudo ./test/wacom-test-env.sh`. And pytest is of course part of `meson test` but that isn't that convenient for single-test debugging.

> don't actually get applied to the virtual device that gets created

If I read your description directly, this is a misunderstanding on how the test suite works. The test suite does create a uinput device which will be visible to the rest of userspace (and picked up by the wacom driver, something we should fix, see #269 ). But the test suite, through the `gwacom` helper library, creates a separate context of the core driver bits - and the options merely change how that context processes files. It's like running two separate `grep` against the same input file - they don't influence each other's output.

Greenscreener:
I see. So if I understand correctly, this applies to the events that reach userspace too and they are going to be different than the ones that are available in the code and these ones are the only that are correct. 

I thought the pytest code did some magic to have the device in X as a normal device, but I now realise that that would require X to restart each time the source code of the driver would change, which is what you're trying to avoid.

whot:
yep, that install/restart cycle was what I was trying to avoid with that whole libgwacom work - makes the driver more testable and easier to reproduce. With #269 you'll no longer see the events in X itself, the test devices should be rejected by default.

adrianbienias:
@Greenscreener thanks for working on this! I'm facing the same issue with jumpy pan movements and it's a really annoying experience to use Wacom tablet on Linux because of that.
Fingers crossed for merging this PR.

Greenscreener:
I finally removed all the comments with `convertAxes` and removed the function, since it is unused and its functionality has been replaced with `convertAxesM`. I also replaced the test for it with a similar test for `convertAxesM`. 

Just a quick question (@whot?), is it possible to include external functions (like the ones I'm using to interact with valuators) in the C test suite? I'm currently getting errors, even though the same functions are used in the same file outside of the test. (Also, would you mind telling me in which file these functions are located (which file to include)? I found this [repo](https://github.com/whot/xserver-pointer-accel) so I could see which functions to use, but that's not the original source)

On another note, I'm currently investigating why the pytest doesn't work. It seems it doesn't kick into panscroll mode at all and sends events just as if the setting wasn't enabled. Is the line `opts["Button"] = "2 pan"` correct for setting this configuration option? The Button config is somewhat special, so I don't have an example to go by.

src/x11/xf86Wacom.c
```diff
@@ -523,22 +511,39 @@ void wcmInitAxis(WacomDevicePtr priv, enum WacomAxisType type,
...
-	                       Absolute);
+			label,
+			min, max, res, min_res, max_res,
+			Absolute);
+
+	if (type == WACOM_AXIS_SCROLL_X) {
+		SetScrollValuator(pInfo->dev, index, SCROLL_TYPE_HORIZONTAL, 0xffff, 0);
+	} else if (type == WACOM_AXIS_SCROLL_Y) {
+		SetScrollValuator(pInfo->dev, index, SCROLL_TYPE_VERTICAL, 0xffff, 0);
+	}
```
whot, line 545: no need for braces here (and I think they should be on a separate line anyway (much to my annoyance :)

src/x11/xf86Wacom.c
```diff
@@ -523,22 +511,39 @@ void wcmInitAxis(WacomDevicePtr priv, enum WacomAxisType type,
...
 	InitValuatorAxisStruct(pInfo->dev, index,
-	                       label,
-	                       min, max, res, min_res, max_res,
-	                       Absolute);
+			label,
+			min, max, res, min_res, max_res,
+			Absolute);
+
+	if (type == WACOM_AXIS_SCROLL_X) {
+		SetScrollValuator(pInfo->dev, index, SCROLL_TYPE_HORIZONTAL, 0xffff, 0);
```
whot, line 542: I don't think that's quite what you want, is it? `0xffff` is the increment,i.e. the amount when we consider a scroll motion a logical wheel click, this value seems a bit high. I'd pick something smaller and scale to that, it's not like we need 32-bit precision for what is usually a few dozen pixels of scrolling.

test/test_wacom.py
```diff
@@ -241,3 +242,67 @@ def axval(axis: str) -> int:
...
+
+def test_scroll(mainloop, opts):
+	"""
+	Check panscrolling works correctly
+	"""
+	dev = Device.from_name("PTH660", "Pen")
+	opts["Button"] = "2 pan"
+	opts["PanScrollThreshold"] = "150"
+
+
```
whot, line 255: `black` should complain about this extra line here :) and a few other style issues in this file

test/test_wacom.py
```diff
@@ -241,3 +242,67 @@ def axval(axis: str) -> int:
...
+		Sev("SYN_REPORT", 0),
+		]
+
+	press_button2 = [
+		Sev("BTN_STYLUS2", 1),
+		Sev("SYN_REPORT", 0),
+		]
+
+	touchdown_pen = [
+		Sev("BTN_STYLUS", 1),
```
whot, line 274: `BTN_STYLUS` is the button on the stylus, touch down is `BTN_TOUCH`

pytest
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+sudo --preserve-env=GI_TYPELIB_PATH LD_LIBRARY_PATH=$LD_LIBRARY_PATH pytest $@
```
whot, line 2: this file needs to be dropped. Better option is now:
```
$ sudo ./test/wacom-test-env.sh
root@localhost $ pytest
```

whot:
> is it possible to include external functions (like the ones I'm using to interact with valuators) in the C test suite? I'm currently getting errors, even though the same functions are used in the same file outside of the test. (Also, would you mind telling me in which file these functions are located (which file to include)? I found this [repo](https://github.com/whot/xserver-pointer-accel) so I could see which functions to use, but that's not the original source)

These functions are part of the X server's API, see [dix/inpututils.c](https://gitlab.freedesktop.org/xorg/xserver/-/tree/master/dix/inpututils.c). But that makes it a bit of a problem, because including those functions is nontrivial and quite messy (look at the git history of the `test` directory, we used to have this as `FAKE_SYMBOL` iirc). But at least these are self-contained so the best option is to copy/paste those functions into our `if TEST` condition so they're only built for the tests.

src/x11/xf86Wacom.c
```diff
@@ -388,68 +385,59 @@ convertAxes(const WacomAxisData *axes, int *first_out, int *num_out, int valuato
...
+			case WACOM_AXIS_RING2: pos = 8; break;
+					       break;
+			default:
+					       abort();
 		}
 
-		first = min(first, pos);
-		last = max(last, pos);
-		valuators[pos] = value;
+		valuator_mask_set(mask, pos, value);
```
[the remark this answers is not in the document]
Greenscreener: I'll do a proper history rewrite/squash before merging. The commit history is very messy.

src/x11/xf86Wacom.c
Greenscreener: This was originally [your](https://github.com/linuxwacom/xf86-input-wacom/pull/222#issuecomment-1027505491) suggestion (although I'm not sure how concrete of a suggestion it was) and I just picked it because I didn't see any reason why not. The value for this variable is somewhat arbitrary, however it puts an upper limit to the user configurable value of `PanscrollThreshold`, if it is larger than our increment and the delta is sufficiently small (but not zero!) the result will get rounded down to zero. This creates a sort of "dead-zone", that makes it only possible to scroll if reaching a minimum speed. This might also make it outright impossible to scroll on a theoretical tablet with a really high polling rate. I'm not sure, what is the highest practical value for `PanscrollThreshold`, I personally have 150, but I'm don't know what someone else might want to use. From playing around with it, 0xffff, or 65535, feels actually quite reasonable. 0xfff, or 4096, might be also fine, but a `PanscrollThreshold` value of 5000, even if somewhat impractical, might still get some use in some specific use case.

src/x11/xf86Wacom.c
```diff
@@ -388,68 +385,59 @@ convertAxes(const WacomAxisData *axes, int *first_out, int *num_out, int valuato
...
 {
 	InputInfoPtr pInfo = priv->frontend;
-	int valuators[7] = {0};
-	int first_val, num_vals;
-
-	convertAxes(axes, &first_val, &num_vals, valuators);
-
-	xf86PostProximityEventP(pInfo->dev, is_proximity_in, first_val, num_vals, valuators);
+	
+	ValuatorMask *mask = valuator_mask_new(9);
```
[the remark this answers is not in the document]
Greenscreener: Resolved in 21f2c20.

src/x11/xf86Wacom.c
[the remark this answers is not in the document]
Greenscreener: This is what the branch is going to look like: https://github.com/Greenscreener/xf86-input-wacom/tree/new-master

src/wcmConfig.c
```diff
@@ -1008,9 +1014,9 @@ Bool wcmDevInit(WacomDevicePtr priv)
 	if (priv->common->wcmModel->DetectConfig)
 		priv->common->wcmModel->DetectConfig (priv);
 
-	nbaxes = priv->naxes;       /* X, Y, Pressure, Tilt-X, Tilt-Y, Wheel */
-	if (!nbaxes || nbaxes > 7)
-		nbaxes = priv->naxes = 7;
+	nbaxes = priv->naxes;       /* X, Y, Pressure, Tilt-X, Tilt-Y, Wheel, Scroll-X, Scroll-Y */
+	if (!nbaxes || nbaxes > 9)
+		nbaxes = priv->naxes = 9;
```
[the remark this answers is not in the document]
whot: > If this works, it might be good to add this as a separate commit prior to everything else?

yes please! And thanks for finding this!

src/x11/xf86Wacom.c
whot: huh, indeed, it was my example. Sorry about that :) Let me think about this "loudly": we're dealing with Wacom devices here in device-resolution units (not mm) and they typically have a resolution around 100 or 200 units/mm. The `PanscrollThreshold` defaults to 1300 or 2600 or 13mm of movement according to the man page, IOW moving 13mm on the tablet generates one scroll click.

So let's assume the scroll increment is 0xffff, corresponding to those 13mm. That means we have ~5000 units per mm available as precision, which is... 25 times the current resolution of a device. That does seem somewhat excessive :) On the other hand, it's really just a number so maybe we shouldn't care too much about it. In XWayland the value is simply `1.0` and then everything is scaled into that range. So come to think of it - maybe 1.0 is the best option here, easy to read and scale into.

Having said all that - this value **must** be completely independent of the device units, it must not correspond to device units or anything because that'll just trigger bugs and hidden dependencies on the device. That increment value is simply a coordinate system without specific meaning, it only provides the *granularity* but not the *distance*. 

The *distance* is defined by the `PanscrollThreshold`, so we scale from device coordinates into the increment using that threshold. So let's assume increment is 1.0 then that means your outgoing delta is simply `delta/panscroll_threshold`. The value of the threshold doesn't matter because that is in device units - 1300 for the 13mm, or 5000 if you want a 50mm scroll zone. That should just work, right? I think the rounding down you saw was merely the result of an integer division?

Greenscreener: > I think the rounding down you saw was merely the result of an integer division?

It is a result of integer division (or type conversion of a float to integer), either way, the whole pipeline for passing values around, including the `wcmAxisSet` function, `WacomAxisData` struct and `convertAxes`, uses integers which means the whole thing would have to be rewritten to use floats in order to be able to pass non-integer values. As you said, it *really is just a number* and isn't worth this change? I'll at least add a comment. There's no real reason for it to be this high, but it's on the safe side and there's no drawback from it being this high if I'm not missing something.

Greenscreener:
> On another note, I'm currently investigating why the pytest doesn't work. It seems it doesn't kick into panscroll mode at all and sends events just as if the setting wasn't enabled. Is the line opts["Button"] = "2 pan" correct for setting this configuration option? The Button config is somewhat special, so I don't have an example to go by.

Damn, after many more hours of digging through the source code, I think I have the answer. There are two ways of getting configuration options to the driver. Via `xorg.conf` or using `xsetwacom`. The major difference is that the options in `xorg.conf` are a simple key-value store and thus cannot accommodate the complicated values that Button actions use. There's only support for the old button remapping from one number to another using `ButtonN` and an integer value, as you can see here:
https://github.com/linuxwacom/xf86-input-wacom/blob/f2df07587e5ff89a6c5b2246bc468211c31cf43a/src/wcmValidateDevice.c#L910-L915
Now, gwacom and thus the test suite use the same pipeline as `xorg.conf` and so don't support remapping buttons. I believe it would require either reimplementing a large part of `xsetwacom`'s functionality into gwacom or somehow enabling the test suite to call `xsetwacom` scoped to the gwacom-powered driver, which is in both cases currently out of scope for this pull request and I think it should be handled separately. For that reason, I'm dropping the pytest bit from this pull request.

<details>
<summary> Here's the source code for future reference, I don't think it's interesting enough to keep in commit history.  </summary>

```python
def test_scroll(mainloop, opts):
    """
    Check panscrolling works correctly
    """
    dev = Device.from_name("PTH660", "Pen")
    opts["Button2"] = "pan"
    opts["PanScrollThreshold"] = "150"

    prox_in = [
        Sev("ABS_X", 50),
        Sev("ABS_Y", 50),
        Sev("BTN_TOOL_PEN", 1),
        Sev("SYN_REPORT", 0),
    ]

    prox_out = [
        Sev("BTN_TOOL_PEN", 0),
        Sev("SYN_REPORT", 0),
    ]

    press_button2 = [
        Sev("BTN_STYLUS", 1),
        Sev("SYN_REPORT", 0),
    ]

    touchdown_pen = [
        Sev("BTN_TOUCH", 1),
        Sev("SYN_REPORT", 0),
    ]

    move_pen_x = [Sev("ABS_X", 75), Sev("SYN_REPORT", 0)]

    up_pen = [Sev("BTN_TOUCH", 0), Sev("SYN_REPORT", 0)]

    depress_button2 = [Sev("BTN_STYLUS", 0), Sev("SYN_REPORT", 0)]

    monitor = Monitor.new_from_device(dev, opts)

    monitor.write_events(prox_in)
    monitor.write_events(press_button2)
    monitor.write_events(touchdown_pen)  # Pen touchdown
    monitor.write_events(move_pen_x)  # Move pen 25% towards positive x
    monitor.write_events(up_pen)  # Pen up
    monitor.write_events(depress_button2)  # Depress button2
    monitor.write_events(prox_out)

    mainloop.run()
    logger.debug(f"We have some events: {monitor.events}")
```
</details>

src/x11/xf86Wacom.c
Greenscreener: In the new code, it is only used once, so I think it should be fine.

whot:
just for reference, `xsetwacom` is a wrapper around X device properties, just with a wacom-specific UI. you can get the same functionality with the generic `xinput` tool except you just have to know the various details of the properties. see the big `parameters[]` array in `xsetwacom.c`.

The button actions are *really* hard to support via the xorg.conf because they contain references between atoms, and those are created at runtime - by the client. Doing this from within the driver is difficult, error-prone and 10y ago or whenever we decided it's not worth the effort. That is primarily true for complex actions, it's less so for the pan scrolling where it's likely the only action assigned to the button (though technically you could map a button to type "hello" followed by toggling panscroll).

Anyway, we can hack around this in gwacom so we can test it, see #287, with that in place I can run your test case (minor modifications) and get pan scroll events emitted, see the branch [here](https://github.com/whot/xf86-input-wacom/pull/new/wip/smooth-panscrolling)

whot:
Apologies this took me so long, things are.... busy...

Anyway, I finally found the time to test this and the pan scrolling does work as expected, it's nice and smooth and equivalent to touchscreen/touchpad scrolling as one would expect. Nice work!

The one thing I am *really* worried about however is that some of the axes change position (see  [my comment](https://github.com/linuxwacom/xf86-input-wacom/pull/222/commits/48b525b79133c5a0d566f5b06ed23726d0242d86#r797207741)). This **will** break some applications that have hardcoded the axis numbers for decades and we won't notice for quite a while, and then fixing it will become even harder.

Doubly so because the two valuators that shift are the artpen rotation and the airbrush slider - both devices that a) most people don't have (including me for example) and thus won't see testing and b) users with those devices are most likely to use software that hardcodes the valuators (e.g. proprietary software that cannot be patched easily).

Now, regarding your comment that the valuator 8 only gets initialized sometimes - we can work around this. It's always uninitialized  on pens [1] so we can make it dual-use - scroll if `IsPen()` otherwise dualring if `IsPad()`. This should be a relatively relatively easy to do (related: #288) and provide the same functionality without breaking anything.


[1] And IIRC the Cintiq 24 was the only one with two rings I *think* that was discontinued in 2015 or so. @pinglinux or @jigpu may remember

Greenscreener:
Hello, 
I am also somewhat busy, but here's the changes you suggested. 
I hope to get around to checking out the tests too soon.

On an unrelated note, for some reason, when I build the repository with meson, which is what I've been doing until now, X crashes immediately after trying to load the shared library. The last two lines in the log are always:
```
[  1455.766] (II) LoadModule: "wacom"
[  1455.767] (II) Loading /usr/lib64/xorg/modules/input/wacom_drv.so
```
After that X just dies with no other explanation.

Weirdly, it workes just fine with the snippet from [here](https://github.com/linuxwacom/xf86-input-wacom/wiki/Building-The-Driver#building-with-autotools).

Greenscreener:
> when I build the repository with meson, which is what I've been doing until now, X crashes immediately after trying to load the shared library

This happens, somewhat understandably, with the tests and gwacom too. It spits out this singular line:
```
==18854==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
```
The same message gets written to the output when running X using startx while the tablet is connected. (again only when building using meson)

Greenscreener:
I know I haven't been awfully fast with implementing these changes, but since it has been almost a month since the last activity here, I thought I might as well jump in with a quick and friendly ping, just so it isn't forgotten entirely.

Pinglinux:
Hi @Greenscreener Peter is taking time off. He'll be back soon. Thank you for your contribution and patience!

whot:
Closing this one. The patches were partially factored out (#296, #297) and the rest was squashed together to apply on top of current master in #298. That PR is effectively the same code as here.

PR #225, opened 2022-02-02
Change our internal test suite to something more maintainable (and extensible)

whot:
The `fake-symbols.c` approach has long been annoying - it was nice to have a few test cases 10y ago bug we're barely making use of it otherwise. We have a lot of little helper functions that could do with a test (e.g. see #224).

This rips out the old `wacom-tests` and replaces it with an ELF section approach [1]:
- tests are behind an `#ifdef ENABLE_TESTS`, directly in the source file
- test cases use a macro that places them into a custom ELF section
- a simple function iterates through that section to call all tests
- custom test driver .so is compiled
- the `wacom-tests` binary simply `dlopen()`s and `dlsym()`s  that function and calls it

The result is that we can have the tests in sito without having to wrap every single X server API. 

This is meson only because autotools is a pain.

Sits on top of #224 and provides the test for that, after revamping the whole lot

[1] Also used by e.g. libinput and pipewire for test discovery. And weson if it's still there, that's where I got the idea from


src/wcmConfig.c
```diff
@@ -1079,4 +1083,162 @@ void wcmDevClose(WacomDevicePtr priv)
...
+
+	common.wcmMaxX = maxx;
+	common.wcmMaxY = maxy;
+	common.wcmResolX = xres;
+	common.wcmResolY = yres;
+
+	wcmInitialToolSize(&priv);
+
+	assert(priv.topX == minx);
+	assert(priv.topY == minx);
```
jigpu, line 1118: This line has a typo on it that I was going to fix in a separate pull request. You can ignore it for now, but I thought I'd mention it.

src/wcmConfig.c
```diff
@@ -1079,4 +1083,162 @@ void wcmDevClose(WacomDevicePtr priv)
...
+
+	common.wcmMaxTouchX = maxx;
+	common.wcmMaxTouchY = maxy;
+	common.wcmTouchResolX = xres;
+	common.wcmTouchResolY = yres;
+
+	wcmInitialToolSize(&priv);
+
+	assert(priv.topX == minx);
+	assert(priv.topY == minx);
```
jigpu, line 1138: Ditto

jigpu:
I like what this pull request does! I think we can change the unit tests inside xsetwacom to act the same -- it uses a similar `ifdef` block at the end of `xsetwacom.c` that should be easily adapted to this same scheme.

src/wcmConfig.c
whot: oooh, an almost 11y old bug! congratulations! :)

Fixed in a follow-up commit, thanks!

src/wcmConfig.c
whot: fixed too

whot:
Good point, xsetwacom now uses the same test suite. Took all of 10 min :)

whot:
Ok, a flurry of commits here:
- xsetwacom uses that test suite too (needed a few revisions to sort out the autotools build)
- renamed from `src/wcmTest.c` to `tests/wacom-test-suite.c` which (now that it's used by both) is better, IMO

jigpu:
I'm getting a build failure with meson on my system:
~~~
$ ninja -C out test
ninja: Entering directory `out'
[46/51] Generating wacom-0.99.gir with a custom command
g-ir-scanner: link: gcc -pthread -o [...]xf86-input-wacom/out/tmp-introspects880iz8d/wacom-0.99 [...]xf86-input-wacom/out/tmp-introspects880iz8d/wacom-0.99.o -L. -Wl,-rpath,. -Wl,--no-as-needed -L[...]xf86-input-wacom/out/ -Wl,-rpath,[...]xf86-input-wacom/out/ -lgwacom -lm -lglib-2.0 -lgobject-2.0 -lgio-2.0 -levdev -lgirepository-1.0 -lgio-2.0 -lgobject-2.0 -Wl,--export-dynamic -lgmodule-2.0 -pthread -lglib-2.0 -lglib-2.0
[49/51] Linking target wacom-tests
FAILED: wacom-tests 
cc  -o wacom-tests wacom-tests.p/test_wacom-tests.c.o -Wl,--as-needed -Wl,--no-undefined '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,[...]xf86-input-wacom/out/ -Wl,--start-group wacom_drv_test.so -Wl,--end-group
/usr/bin/ld: wacom-tests.p/test_wacom-tests.c.o: in function `main':
[...]xf86-input-wacom/out/../test/wacom-tests.c:29: undefined reference to `dlopen'
/usr/bin/ld: [...]xf86-input-wacom/out/../test/wacom-tests.c:31: undefined reference to `dlerror'
/usr/bin/ld: [...]xf86-input-wacom/out/../test/wacom-tests.c:35: undefined reference to `dlsym'
/usr/bin/ld: [...]xf86-input-wacom/out/../test/wacom-tests.c:38: undefined reference to `dlerror'
collect2: error: ld returned 1 exit status
[50/51] Generating force GIR build with a custom command (wrapped by meson to capture output)
ninja: build stopped: subcommand failed.
~~~

It looks like it can be fixed by adding something a dependency like `dep_dl = meson.get_compiler('c').find_library('dl', required : true)` to the test suite. Of course, why its working fine for you and Github's automated tests are beyond me...

Also: would you mind me trying my hand at getting fixing up autotools to work with the new style? My brain is still wired for it and was able to fix up `make check` for xsetwacom easily enough...

whot:
`dl` is linked in now, so hopefully this works.

> Also: would you mind me trying my hand at getting fixing up autotools to work with the new style? My brain is still wired for it and was able to fix up make check for xsetwacom easily enough... Plus I was kinda hoping we could drop autotools at some point in the foreseeable future.

fwiw, the xsetwacom part is the easy one, the harder one is the other bit. You can have a go but tbh, I'd advise against it, you spend less time retraining yourself to use ninja than you'd be wasting on autotools here (with the ongoing maintenance effort of that too).

I have [this `n` script](https://gist.github.com/whot/e9864263ec9ef3aebe43520be5855058) in my `$PATH`. This started because I was sick of typing `ninja -C build` and expanded into cargo + make support. so now I just type `n` and it'll automatically build with cargo or meson, or fall back to make, depending what's around in the repo. Then it's just down to remembering to run `n test` (`ninja`) vs `n check` (`make check`).

jigpu:
> `dl` is linked in now, so hopefully this works.
>

Yup, works fine now.

> > Also: would you mind me trying my hand at getting fixing up autotools to work with the new style? My brain is still wired for it and was able to fix up make check for xsetwacom easily enough... Plus I was kinda hoping we could drop autotools at some point in the foreseeable future.
> 
> fwiw, the xsetwacom part is the easy one, the harder one is the other bit. You can have a go but tbh, I'd advise against it, you spend less time retraining yourself to use ninja than you'd be wasting on autotools here (with the ongoing maintenance effort of that too).
> 
> I have [this `n` script](https://gist.github.com/whot/e9864263ec9ef3aebe43520be5855058) in my `$PATH`. This started because I was sick of typing `ninja -C build` and expanded into cargo + make support. so now I just type `n` and it'll automatically build with cargo or meson, or fall back to make, depending what's around in the repo. Then it's just down to remembering to run `n test` (`ninja`) vs `n check` (`make check`).

Its less a matter of learning to use meson/ninja than being confident that the tests work regardless of how warmed up my brain is that day (i.e., did I check / distcheck out of habit?). That `n` script is a good idea though -- I think I'll have to swipe it :)

That said, the deed has already been done and I've pushed the changes to re-enable the tests on autoconf to a branch in my repo: https://github.com/jigpu/xf86-input-wacom/commits/test-rework. You'll find two commits with the prefix `UPDATE` that can be squashed into your changes if you're happy with how they look. I've also tacked on two additional changes to the end of the branch that I noticed along the way.

whot:
sigh, ok, this was frustrating. I changed your patch from `check_PROGRAMS` to `noinst_PROGRAMS` so we build the tests during a normal build run. automake otherwise only builds them during `make check` which is frustrating when you're developing (you suddenly deal with compilation errors when you think you're done).

Of course this now breaks in the test suite because of autotool's broken `rint` check - we call `make CFLAGS=$CFLAGS` with our custom cflags but that now fails because of `-Werror` since we don't call some of the xsetwacom functions. I've fixed this with an extra commit that marks those as `static inline` so we don't have to hack around this in the build system.

Then the build failed because autotools still defaults to c90 instead of c99. Fixed that.

Then it failed because the `static inline` hack didn't work on clang because of course it doesn't.

So I dropped all these, went back to `check_PROGRAMS` and any frustration with compilation errors during `make check` are now squarely in the hands of those that want to keep using autotools :P

Your patches are squashed in/put on top, so this is ready to go now. Thanks.

jigpu:
Oof. Reverting back to `check_` and making it a second-class (but still "working") citizen to the meson path works for me. Apologies for the unintentional frustration :grimacing:

PR #231, opened 2022-02-04
autotools: default to C99

whot:
Because some things we got added to C in the 90s [1] are quite nice to have, or at least not a reason to throw compilation errors.

[1] yes, the 90s. That's how far ahead of the curve we are.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -36,6 +36,8 @@ AM_INIT_AUTOMAKE([foreign dist-bzip2 no-dist-gzip subdir-objects])
 AM_MAINTAINER_MODE([enable])
 AC_USE_SYSTEM_EXTENSIONS
 
+AC_PROG_CC_STDC
+
 # Initialize libtool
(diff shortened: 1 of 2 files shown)


jigpu:
One of the patches on the branch I'm deferring until after this coming release is chock-full of these fancy "new" C99 inline declarations and commit message snark. I'm all for this.

Autotools 2.70 apparently deprecates `AC_PROG_CC_STDC` in favor of `AC_PROG_CC` magically preferring the latest standard. Of course, `AC_PROG_CC` does no such thing in earlier autotools releases, so I don't see any way to avoid the deprecation warning. Sigh.

whot:
sigh. I guess we could use `AC_AUTOCONF_VERSION` to have this conditional on the autoconf version. But right now, I think we're good to use a deprecated autoconf macro - let's worry about this when they actually remove the macro and it stops working. Hopefully by then we have dropped autotools support.

Issue #226, opened 2022-02-02
Absolute values passed to valuators in relative mode


Greenscreener:
While working on #222, I found a bug which I originally assumed to be related to my changes, but it's actually present in master. 

When in relative mode, it seems that absolute values are passed to the valuator even in relative mode, which makes the cursor fly off to the bottom edge of the screen. I bisected the problem and it seems to be introduced in 58a931bc21d1e78629a0873beb221252d065fdca.

Greenscreener:
A workaround that seems to work is removing these two [lines]( https://github.com/linuxwacom/xf86-input-wacom/blob/master/src/wcmCommon.c#L700-L701). I haven't been able to find out why or how as of yet.

whot:
Referencing #221, need to fix this before the release. Good find, thank.

whot:
This doesn't appear to happen in the core driver:
```
$ sudo ./builddir/wacom-record  --options "Mode=Relative" /dev/input/event4
    - source: 0
      event: motion
      axes: { x:    13, y:     7, pressure:    0, tilt: [  0,  0], rotation:   0, throttle:   0, wheel: -108, rings: [  0,   0] }
    - source: 0
      event: motion
      axes: { x:    15, y:     9, pressure:    0, tilt: [  0,  0], rotation:   0, throttle:   0, wheel: -108, rings: [  0,   0] }
    - source: 0
      event: motion
      axes: { x:    19, y:     9, pressure:    0, tilt: [  0,  0], rotation:   0, throttle:   0, wheel: -108, rings: [  0,   0] }
    - source: 0
      event: motion
      axes: { x:    20, y:     9, pressure:    0, tilt: [  0,  1], rotation:   0, throttle:   0, wheel: -108, rings: [  0,   0] }
    - source: 0
      event: proximity
      proximity-in: false
      axes: { x:     0, y:     0, pressure:    0, tilt: [ 30, 30], rotation:   0, throttle:   0, wheel: -900, rings: [  0,   0] }
```

I ran the xorg driver with master from today (be159bcd06c39ee1839efa66b732c810b6afe186) and I'm unable to reproduce with that too. It's a bit suspicious that we're posting all zeros (the mask is set for those when it probably shouldn't be) but i think that's what the driver has always done (confirmed with wacom 0.40.0 at least).

jigpu:
[...]

PR #241, opened 2022-02-09
Add a driver test suite through pytest

whot:
This adds a python-based test suite (hooked up through pytest) so we can test the driver internals. This won't work for the X11 specific bits but with the gwacom library we can easily test the rest of the driver.

Needed to run locally, for `pytest` to run against meson `builddir`:
```
$ export GI_TYPELIB_PATH="$PWD/builddir:$GI_TYPELIB_PATH"
$ export LD_LIBRARY_PATH="$PWD/builddir:$LD_LIBRARY_PATH"
```

cc @jigpu, an example test for your fix to #226 is included here

edit: `pytest` needs to be run as root since we're going through uinput and are opening event nodes

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,15 @@
+on: [ push, pull_request ]
+
+jobs:
+  python-black:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+        with:
+           python-version: '3.x'
+      - run: pip install black
+      - run: black test/
+      - run: git diff --exit-code || (echo "Please run black to reformat Python files" && false)
+
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,8 +5,8 @@ env:
   UBUNTU_PACKAGES: |
     xutils-dev xserver-xorg-dev libx11-dev libxi-dev libxrandr-dev libxinerama-dev libudev-dev
     libgirepository1.0-dev libevdev-dev
-    python3-pip
-  PIP_PACKAGES: meson ninja
+    python3-pip python3-gi python3-pytest
+  PIP_PACKAGES: meson ninja libevdev pytest pyyaml attrs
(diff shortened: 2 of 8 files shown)


whot:
cc @Pinglinux 

Gonna provide a bit more details here because I think this is quite the game changer. With the libgwacom library we get the GIR bindings that can be used from Python (or other languages). Since the actual surface of the core driver API is quite small - mostly options in, log messages and events out, we can wrap that into a few helper classes  in Python  [1], specifically here the `Device` and `Monitor` class which binds a few things together. The `Monitor` class is responsible for collecting events from the driver into a simple list.

We have a set of yaml files (just one for now) that describe the various devices, the test then 
- loads a device from a yaml file
- sets up the uinput device and monitor
- writes events to the uinput device
- checks that the events we got out match what we expect

A shortened version of the proximity test:
```python
    dev = Device.from_name("PTH660", "Pen")  # this loads the Pen event node from a PTH660
    monitor = Monitor.new_from_device(dev, opts)  # hooks everything up
    monitor.write_events(...) # this is where we write the evdev events
    assert isinstance(monitor.events[0], Proximity)
    assert monitor.events[0].is_prox_in
    assert monitor.events[0].axes.x == somevalue
```

And because it's based on `pytest` that's the only command you need to run the tests (well, and setting up the two environment variables).

Now, long-term we may want to include default sequences into the yaml files (I'm doing this for libinput so "touch down" has the right sequence of bits for each test device) but even with just this we can go pretty wild. Especially compared to the previous approach. All this was written in a day, so we can assume we can think of more interesting bits in the future.

The only major limitation is: we're going though uinput which limits what we can do in a container (at least on github, gitlab.freedesktop.org would allow us to run these in KVMs).

[1] I'm using [attr](https://www.attrs.org/en/stable/examples.html) a lot which is a great python lib

whot:
Just before I merge this a caveat to point out: `pip` installs in the user's home directory by default but we need to run the tests as root.`pip install && sudo pytest` will have interesting failures because a package that's installed may not be findable by root.

The general recommendation is thus either `pip` installing as root or using the system packages for the various python bits.

PR #249, opened 2022-02-14
Correct typo in calculation of relative-mode non-pad RING2 values

jigpu:
The current value needs to be read from the RING2 axis, not the plain
RING axis.

Fixes: 58a931bc21d1 ("Abstract the event interface to pass a struct with axis data around")


whot (approved):

whot:
Thanks, merged. This should really have a test case in #241, RING2 is not exactly an easy one to test otherwise.

PR #253, opened 2022-02-17
meson: Fix fuzzinterface build option

jigpu:
Trying to enable this option results in the following configuration
error:

    meson.build:117:10: ERROR: Configuration set requires 2 arguments.


whot (approved):
oops

Issue #257, opened 2022-03-15
Allow negative panscrolling speed pwease UwU


MikuChan03:
Heya!

I'm in a bit of a bad mood so I'll spare you the whole Miku-Chan bit. Anyways, I'm one of the borderline insane people that like "unnatural" scroll, as libinput puts it, and want the screen contents to move upwards when my pen moves upwards. This of course works nicely with setting a negative panscrolling speed. Not wanting to do that at every reboot, I put it in a x11 config file. This also works tremendously, when setting a positive value that is. The Xserver is alright with me setting the speed to for example "1234". If I were to set it to "-1234" though, it goes like, awww, poor guy, lemme just fix this silly input mistake you make, and reverts it back to 1300 or 2600. I am very appreciative of you guys looking out to me to this extent, and I do I feel like I should at least send you guys a christmas card or whatever. Due to countless cerebral insufficientcies of mine, at times I really truly honestly would prefer a negative panscrolling speed, right from the getgo. So if one of you nice senpais would be so kind as to entertain this insufferable foolishness of mine, I would be boundlessly, everlastingly grateful.

Thank!

It goes like this (viewer discretion is advised):
```
[  2683.739] (--) wacom: GAOMON Gaomon Tablet Pen stylus: using pressure threshold of 851 for button 1
[  2683.739] (**) wacom: GAOMON Gaomon Tablet Pen stylus: panscroll is -1234
[  2683.739] (**) wacom: GAOMON Gaomon Tablet Pen stylus: panscroll modified to 2600
[  2683.739] (--) wacom: GAOMON Gaomon Tablet Pen stylus: maxX=33020 maxY=20320 maxZ=8191 resX=200000 resY=200000  tilt=enabled
```

Also, if Senpai Hutterer where to read this. Thanks for  believing in me, man. I asked you a while back how to set relative cursor movement with libinput and you said that while it sadly doesn't work, I can implement it if I want to. While I didn't do that, I actually added native support for some newer xppen tablet to digimend and sent the patch of the mr. Kosina and mr. Tissoire. They haven't had a look at it yet, but it's nothing to crazy and I have hopes for it getting accepted. So thanks for your encouragement! Also thanks for implementing panscroll at all.

MikuChan03:
I really don't want to bother anyone, but all I'm suggesting is that we might go:
```c
wcmLog(priv, W_CONFIG, "panscroll is %d\n", common->wcmPanscrollThreshold);
[...]

PR #260, opened 2022-04-13
Allow negative panscroll threshold for inverted scrolling

jigpu:
The driver's "pan" feature allows you to send scroll events by holding down
a defined stylus button and dragging the pen. Dragging the pen from top
to bottom will cause the driver to send scroll-up events, and vice-versa.
This simulates physically dragging the page under the cursor and works
particularly well with display tablets.

The rate at which scroll events are sent is controlled by comparing the
distance the pen has moved "down" to a defined threshold value. If the
distance exceeds the threshold a scroll-up event is sent; if it exceeds
the the threshold * -1 a scroll-down event is sent.

This commit allows the driver to accept negative threshold values to
produce an inverted scrolling behavior (i.e. dragging the pen from top
to bottom causes scroll-down events instead of scroll-up). To do this
we just relax the checks performed at device configuration and property
setting (affecting the "PanScrollThreshold" driver option and "Wacom
Panscroll Threshold" Xinput property, respectively).

Link: https://github.com/linuxwacom/xf86-input-wacom/issues/257

### DIFF
--- a/src/x11/xf86WacomProperties.c
+++ b/src/x11/xf86WacomProperties.c
@@ -931,16 +931,21 @@ static int wcmSetProperty(DeviceIntPtr dev, Atom property, XIPropertyValuePtr pr
 			common->wcmPressureRecalibration = values[0];
 	} else if (property == prop_panscroll_threshold)
 	{
-		CARD32 *values = (CARD32*)prop->data;
+		INT32 *values = (INT32*)prop->data;
 
 		if (prop->size != 1 || prop->format != 32)
 			return BadValue;
 
-		if (values[0] <= 0)
-			return BadValue;
-
 		if (IsTouch(priv))
 			return BadMatch;
+
+		/* Reset to default if set to 0 */
+		if (values[0] == 0) {
+			values[0] = common->wcmResolY * 13 / 1000; /* 13mm */
+		}
+		if (values[0] == 0) {
+			values[0] = 1000;
+		}
 
 		if (!checkonly)
 			common->wcmPanscrollThreshold = values[0];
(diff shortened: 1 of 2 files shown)


MikuChan03:
Big facts.

src/x11/xf86WacomProperties.c
```diff
@@ -931,17 +931,22 @@ static int wcmSetProperty(DeviceIntPtr dev, Atom property, XIPropertyValuePtr pr
 			common->wcmPressureRecalibration = values[0];
 	} else if (property == prop_panscroll_threshold)
 	{
-		CARD32 *values = (CARD32*)prop->data;
+		INT32 *values = (INT32*)prop->data;
```
whot, line 934: huh, interesting, it was already an `INT32` anyway, so this isn't even an API break.

src/x11/xf86WacomProperties.c
```diff
@@ -931,17 +931,22 @@ static int wcmSetProperty(DeviceIntPtr dev, Atom property, XIPropertyValuePtr pr
...
 		if (IsTouch(priv))
 			return BadMatch;
 
+		/* Reset to default if set to 0 */
+		if (values[0] == 0) {
+			values[0] = common->wcmResolY * 13 / 1000; /* 13mm */
+		}
+		if (values[0] == 0) {
+			values[0] = 1000;
+		}
```
whot, line 948: 13 is a rather magic number to pick for this, any particular reason? I'm not disagreeing with it, but it would be good to know if there's more behind it than "seems good enough".

jigpu: Code was copied to match how the value is initialized on device connect. I probably should have refactored somehow, but laziness got the better of me :sweat_smile: The value 1000 was pretty much just a "right order of magnitude" choice that shouldn't normally be necessary (we "should" always have the kernel resolution) but avoids the possibility of division by zero.

https://github.com/linuxwacom/xf86-input-wacom/blob/742c3e4494bf89301031ba6a627a7f5d5f49556f/src/wcmCommon.c#L1565-L1572

whot: oh, indeed, the 13mm is documented in the man page. good enough for me, thanks!

Issue #168, opened 2021-03-13
S-pen stylus not recognized (Ubuntu 20.10)


waitlamp:
Hello,
I have filed an issue in wacom-hid-descriptors [#147](https://github.com/linuxwacom/wacom-hid-descriptors/issues/147)
On ubuntu 20.10, Samsung Chromebook Pro s-pen can use, 
but xsetwacom cannot detect any wacom devices, GNOME either.
Is there any way to use xf86-input-wacom driver? 
Any help is appreciated.
thanks!

here is sysinfo
[sysinfo.UwjeIiKZUM.tar.gz](https://github.com/linuxwacom/wacom-hid-descriptors/files/6129840/sysinfo.UwjeIiKZUM.tar.gz)

whot:
```
$ cat > /etc/X11/xorg.conf.d/99-wacom-driver.conf <<EOF
Section "InputClass"
        Identifier "Wacom USB tablet class"
        MatchUSBID "2d1f:*"
        MatchDevicePath "/dev/input/event*"
        MatchIsTablet "true"
        Driver "wacom"
EndSection
EOF
```
should do the trick, based on the USB id from the wacom-hid-descriptor bug

waitlamp:
It work.
`xsetwacom --list` now list four devices. (some devices' name are same.)
```
$ xsetwacom --list
ACPI0C50:00 2D1F:0163 stylus    	id: 11	type: STYLUS    
ACPI0C50:00 2D1F:0163 stylus    	id: 12	type: STYLUS    
ACPI0C50:00 2D1F:0163 eraser    	id: 14	type: ERASER    
ACPI0C50:00 2D1F:0163 eraser    	id: 15	type: ERASER    
```
but GNOME setting still not recognized the stylus.
in Krita, Tablet Tester detected the Pen tip brought near / taken away 
~~but no pressure and X-Y axis , and when drawing, pressure sensitive work~~ 
Add: I have tried many times and a few times I could make X-Y axis and pressure level recognized.
(in most cases, only Pen tip brought near / taken away can be recognized)
And I tried to use touch/mouse and stylus at same time, pen tip caused mouse drifting, also moved mouse X-Y axis.
It seems pen still recognized as mouse when pen tip drawing on screen.

Then, I tried to use `xsetwacom set "11" "Mode" "Absolute"`
(both 12 / 14 / 15, devices have same devices name, I have to use ID instead of name.) 
but problem still exist.
Is a krita problem？

[...]

PR #264, opened 2022-04-22
conf: add Samsung Galaxy Book Pro 360

smilejj91:
Hello.

This request is a conf file patch request to support the wacom device whose vendorID is recognized as 2d1f.

1. Cause of the problem
- Samsung Galaxy Book Pro 360 uses a wacom device, and the vendorID of the device is 2d1f, not 056a.
- So, the device is not recognized by xf86-input-wacom (= xserver-xorg-input-wacom package)
- Looking through the issues in the past, there were many users who faced the same problem as me.

2. Suggest solutions
- I hope the following patch is applied to /usr/share/X11/70-wacom.conf so that the 2d1f vendorID can be recognized as a wacom device.

3. Reference
- https://github.com/linuxwacom/xf86-input-wacom/issues/168

Thank you.

### DIFF
--- a/conf/70-wacom.conf
+++ b/conf/70-wacom.conf
@@ -127,4 +127,11 @@ Section "InputClass"
         Driver "wacom"
 EndSection
 
-
+# Samsung Galaxy Book Pro 360 (Supports S-Pens)
+Section "InputClass"
+        Identifier "Samsung Galaxy Book Pro 360"
+        MatchUSBID "2d1f:*"
+        MatchDevicePath "/dev/input/event*"
+        MatchIsTouchscreen "true"
+        Driver "wacom"
+EndSection


whot:
0x2d1f usually means the firmware is in [android mode](https://github.com/linuxwacom/xf86-input-wacom/wiki/Known-Issues#issue-my-wacom-intuos-ctl-4100-or-ctl-6100-seems-to-be-miscalibrated-and-has-a-dimmed-status-led), see also [this wiki page here](https://github.com/linuxwacom/libwacom/wiki/Troubleshooting#special-cases). IIRC the tablets don't behave the same way when they're in android mode so there may be subtle other bugs awaiting too, better to convince the firmware of the realities of life :)

smilejj91:
Oh, now I see what 0x2d1f means is related to android mode.

By referring to the wiki page you linked, we will apply the appropriate firmware to the PC.

thank you.

PR #265, opened 2022-05-05
Test artpen rotation

whot:
cc @skomra 

A few useful cleanups, then a few necessary cleanups, and finally the small change to test artpen rotation comes through as expected.

Confusingly, artpen rotation is **not** sent through the rotation axis but through the "wheel" axis. rotation is only for cursor/puck rotation. The artpen physical rotation in `ABS_Z` is handled by the driver as wheel because the X11 valuator is shared with the airbrush wheel.

### DIFF
--- a/src/WacomInterface.h
+++ b/src/WacomInterface.h
@@ -58,9 +58,9 @@ enum WacomAxisType {
 	WACOM_AXIS_TILT_Y	= (1 << 4),
 	WACOM_AXIS_STRIP_X	= (1 << 5),
 	WACOM_AXIS_STRIP_Y	= (1 << 6),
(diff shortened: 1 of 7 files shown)


whot:
> Which recording are you referring to in https://github.com/linuxwacom/xf86-input-wacom/commit/df6723ff149e589bb147d44bb0bae78054f849a4?

Not an actual problem at the timeline of the patch but without it the last commit (07b857d9921e3f3ca82ed8d94503799e7c4764f3) won't work. The VID is assigned to the pth660 product id (our test device recording) and then the `IsArtPen` checks don't work because they check for the wacom VID.

Thanks for the review!

Issue #126, opened 2020-05-26
xsetwacom returns 0 even when there's an error


maruseu:
xsetwacom only seems to return a non zero when its run with absolutely no arguments

```
$ xsetwacom --set "Wacom Intuos S Pen stylus" Area 0 0 0 && echo success || echo failure
'Area' requires exactly 4 value(s).
success
```

jigpu:
Looks like there are a few error paths that need to be updated...

AstroLad47:
This is a fairly major task. I started to patch a version for my own use. After fixing get() and set() and still seeing the same problem, I looked deeper and discovered that many (most???) of the functions do not return status.They print to STDERR, but don't return an error code to the caller.

Joshua-Dickens:
This should be fully fixed with the commits above and return error codes properly.

PR #266, opened 2022-05-09
Fix xsetwacom returning 0 even when there's an error

Joshua-Dickens:
xsetwacom was only able to return a failure on a few scenarios but would return a success even when the command would encounter failures.

List, set, get, and some other functions within those all had the potential to fail and would print an error message when said failures occurred but would improperly return a success.
This commit adds the ability to properly report a success(0) or a failure(-1) to the relevant functions including all get and set methods that can be called by get_func and set_func.

Link: https://github.com/linuxwacom/xf86-input-wacom/issues/126
Signed-off-by: Joshua Dickens <joshua.dickens@wacom.com>

### DIFF
--- a/tools/xsetwacom.c
+++ b/tools/xsetwacom.c
@@ -94,8 +94,8 @@ typedef struct _param
 	const unsigned int prop_offset;	/* offset (index) into the property values */
 	const int arg_count;   /* extra number of items after first one */
 	const unsigned int prop_flags;
-	void (*set_func)(Display *dpy, XDevice *dev, struct _param *param, int argc, char **argv); /* handler function, if appropriate */
-	void (*get_func)(Display *dpy, XDevice *dev, struct _param *param, int argc, char **argv); /* handler function for getting, if appropriate */
+	int (*set_func)(Display *dpy, XDevice *dev, struct _param *param, int argc, char **argv); /* handler function, if appropriate */
+	int (*get_func)(Display *dpy, XDevice *dev, struct _param *param, int argc, char **argv); /* handler function for getting, if appropriate */
 
 	/* filled in by get() */
 	char *device_name;
@@ -104,16 +104,16 @@ typedef struct _param
 
 
 /* get_func/set_func calls for special parameters */
-static void map_actions(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void set_mode(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void get_mode(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void get_map(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void set_rotate(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void get_rotate(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void set_xydefault(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void get_all(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void get_param(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
-static void set_output(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int map_actions(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int set_mode(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int get_mode(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int get_map(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int set_rotate(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int get_rotate(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int set_xydefault(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int get_all(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int get_param(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
+static int set_output(Display *dpy, XDevice *dev, param_t *param, int argc, char **argv);
 
 /* NOTE: When removing or changing a parameter name, add to
  * deprecated_parameters.
@@ -955,7 +955,7 @@ static void list_mod(Display *dpy)
 		" See the man page for details.\n");
 }
 
-static void list(Display *dpy, int argc, char **argv)
+static int list(Display *dpy, int argc, char **argv)
 {
 	TRACE("'list' requested.\n");
 	if (argc == 0)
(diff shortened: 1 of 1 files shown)


tools/xsetwacom.c
```diff
@@ -2934,14 +2980,14 @@ int main (int argc, char **argv)
...
 	else if (do_set)
-		set(dpy, argc - optind, &argv[optind]);
+		status = set(dpy, argc - optind, &argv[optind]);
 	else if (do_get)
-		get(dpy, format, argc - optind, &argv[optind]);
+		status = get(dpy, format, argc - optind, &argv[optind]);
 
 	XCloseDisplay(dpy);
-	return 0;
+	return status;
```
whot, line 2990:
```suggestion
	return status == 0 ? 0 : 1;
```

returning `-1` as an exit status is generally not ideal. the exit status is unsigned so -1 ends up being 255 in the shell and 128 and above usually indicate signal-based termination.

A better approach here would be to separate errors (exit status 1) from invalid usage (exit status 2) and return those values accordingly. `EXIT_SUCCESS` and `EXIT_FAILURE` exist in some system header file somewhere already, would be easy to `#define EXIT_INVALID_USAGE 2` and go with that then.

Note that invalid usage exit status 2 is merely one convention and not all programs use it.

whot (changes requested):

tools/xsetwacom.c
Joshua-Dickens: Thanks for the feedback, I'm not very well versed in the exit status meanings so I appreciate the explanation and suggestion I'll be sure to implement the suggested changes.

whot: Actually, found this one which explains it best: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html 

But for us, the only ones that matter are 0, 1 and 2

Pinglinux: > returning -1 as an exit status is generally not ideal. the exit status is unsigned so -1 ends up being 255

If the exit status is unsigned, should we claim the return value as "unsigned char"?

whot: that'd be breaking about 50 years of C convention :)

tools/xsetwacom.c
```diff
@@ -47,6 +47,8 @@
 #define safe_realloc(p, n, s) \
 	((size_t)-1 / (n) < (s) ? NULL : realloc((p), (n)*(s)))
 
+#define EXIT_INVALID_USAGE 2;
```
whot, line 50: hmm, the compiler should warn about this, or at least about a double semicolon when you do this:
```c
   return EXIT_INVALID_USAGE;
```
because it translates to:
```c
   return 2;;
```

As a general rule: *never* terminate a `#define` with a semicolon, unless there's a really good reason for it.

tools/xsetwacom.c
```diff
@@ -969,8 +971,11 @@ static void list(Display *dpy, int argc, char **argv)
 	else if (strcmp(argv[0], "mod") == 0 ||
 		 strcmp(argv[0], "modifiers") == 0)
 		list_mod(dpy);
-	else
+	else{
```
whot, line 974:
```suggestion
	else {
```

tools/xsetwacom.c
```diff
@@ -1442,17 +1447,20 @@ static Bool parse_actions(Display *dpy, int argc, char **argv, unsigned long* da
...
 	unsigned long *data, *btnact_data;
 	Atom type, prop = 0;
 	int format;
 	unsigned long btnact_nitems, bytes_after;
 	unsigned long nitems = 0;
+	int status = EXIT_SUCCESS;
 
 	data = calloc(256, sizeof(unsigned long));
-	if (!parse_actions(dpy, argc, argv, data, &nitems, 256))
+	if (!parse_actions(dpy, argc, argv, data, &nitems, 256)){
```
whot, line 1460:
```suggestion
	if (!parse_actions(dpy, argc, argv, data, &nitems, 256)) {
```

tools/xsetwacom.c
```diff
@@ -1442,17 +1447,20 @@ static Bool parse_actions(Display *dpy, int argc, char **argv, unsigned long* da
...
  */
-static void special_map_property(Display *dpy, XDevice *dev, Atom btnact_prop, unsigned int offset, int argc, char **argv)
+static int special_map_property(Display *dpy, XDevice *dev, Atom btnact_prop, unsigned int offset, int argc, char **argv)
 {
 	unsigned long *data, *btnact_data;
 	Atom type, prop = 0;
 	int format;
 	unsigned long btnact_nitems, bytes_after;
 	unsigned long nitems = 0;
+	int status = EXIT_SUCCESS;
```
whot, line 1457: just a general comment, not necessary to apply in this patch: it's *usually* neater to default to the most common error and then use this pattern:
```c
  int status = EXIT_FAILURE;
  if (blah)
    goto out;
  if (!foo)
    goto out;
  if (bar) {
    status = EXIT_INVALID_USAGE;
  }
  status = EXIT_SUCCESS;
out:
  return status;
```
Reason being that you can use `goto out` as single line in most cases and you don't accidentally return success until you get to the end.

tools/xsetwacom.c
```diff
@@ -2567,39 +2597,48 @@ static Bool set_output_next(Display *dpy, XDevice *dev)
...
 	else
+	{
 		fprintf(stderr, "Unable to find an output '%s'.\n", argv[0]);
+		return EXIT_FAILURE;
+	}
+
+	if(success)
+		return EXIT_SUCCESS;
+	else
+		return EXIT_FAILURE;
```
whot, line 2634:
```suggestion
		return success ? EXIT_SUCCESS : EXIT_FAILURE;
```

tools/xsetwacom.c
```diff
@@ -2626,25 +2665,29 @@ static void get_all(Display *dpy, XDevice *dev, param_t *param, int argc, char *
...
+					success = get_param(dpy, dev, p, 1, &tmpargv);
 				}
 			}
 			else
-				get_param(dpy, dev, p, argc, argv);
+				success = get_param(dpy, dev, p, argc, argv);
 		}
 		p++;
+		if(success == 0)
+			status = EXIT_SUCCESS;
```
whot, line 2676: This always succeeds if at least one parameter succeeds, is this intentional?

tools/xsetwacom.c
```diff
@@ -2934,14 +2983,14 @@ int main (int argc, char **argv)
...
 	else if (do_set)
-		set(dpy, argc - optind, &argv[optind]);
+		status = set(dpy, argc - optind, &argv[optind]);
 	else if (do_get)
-		get(dpy, format, argc - optind, &argv[optind]);
+		status = get(dpy, format, argc - optind, &argv[optind]);
 
 	XCloseDisplay(dpy);
-	return 0;
+	return status == 0 ? 0 : 1;
```
whot, line 2993:
```suggestion
	return status;
```
because status is now set to one of the exit codes, right?

tools/xsetwacom.c
Joshua-Dickens: That's my bad, didn't catch that when writing it.

tools/xsetwacom.c
Joshua-Dickens: Yes now that it can be any of the exit codes putting it back to returning just the status should be proper.

tools/xsetwacom.c
whot: nah, I'm fine with the behaviour, I just want to make sure it's intentional :)

tools/xsetwacom.c
Pinglinux: Ha, that 50 years old specification left enough room for the return value to grow. That's quite robust ;).

whot:
Merged, thanks!

PR #269, opened 2022-05-23
test: ignore test devices, or real devices during test suite runs

whot:
Set an option on all our created uinput devices, and an environment
variable when we're running the test suite. If both of these are set we
process the device, otherwise ignore the device during PreInit.

This stops the driver picking up events from test suite runs
(potentially clicking around on the desktop) and it stops the test suite
from false positives by locally connected devices.


skomra:
Reviewed-by: Aaron Skomra <aaron.skomra@wacom.com>

PR #261, opened 2022-04-18
Don't report fifth valuator unless it exists on tool

skomra:
Fixes #199

Prior to this commit we assumed that all non Art Pens needed
their fifth valuator normalized.

Signed-off-by: Aaron Skomra <aaron.skomra@wacom.com>

PR #270, opened 2022-05-26
 Don't report fifth valuator unless it exists on tool (v2)

whot:
Supersedes #261 

@skomra's patch is identical to #261, only the patch to add the test is different.

Arguably the test should be a separate one but that would require copy/paste of 90% of the test code which is more likely to become an issue than a single condition in the current test that differs based on stylus type, so let's go with that.

### DIFF
--- a/src/wcmCommon.c
+++ b/src/wcmCommon.c
@@ -812,6 +812,7 @@ wcmSendNonPadEvents(WacomDevicePtr priv, const WacomDeviceState *ds,
 }
 
 #define IsArtPen(ds)    (ds->device_id == 0x885 || ds->device_id == 0x804 || ds->device_id == 0x100804)
+#define IsAirBrush(ds)  (ds->device_id == 0x902 || ds->device_id == 0x100902)
 
 static void
 wcmUpdateSerial(WacomDevicePtr priv, unsigned int serial, int id)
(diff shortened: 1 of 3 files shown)


whot:
> Also maybe we should make sure Ping's recent kernel patch isn't inadvertently fixing the problem on your system

we're using uinput here, so the kernel patches don't matter. the evdev device is set up with our specific axes and ranges and we send the events through uinput - no kernel patch is involved here. that's what makes uinput-based testing so nice, you can use the kernel infrastructure (i.e. udev, etc.) but you decouple completely from the actual kernel driver.

> Can you think of what I might be missing to get this error on the fresh system? I thought it might be attrs but I have that installed.

that error is confusing... if it was attr you'd think the error would happen at `import attr`, not when it's actually used. maybe `attr.ib()` local-imports something else? You can try this:
```
$ ./test/wacom-test-env.sh
$ python3
>>> import test
>>> from test import InputId
>>>
```
That would import the test module and then for good measure the class that triggers that error too. If that throws an import error it may at least have more information.

Oh, and - if you `pip install` something as user but then run the tests as root it won't find the package. Maybe that's it?



As for the test case failure.... no idea why this could happen, this isn't host-specific so the result should be the same regardless of distribution/kernel/userspace. `pytest --verbose --log-level=DEBUG` may have some info on what the driver actually does.

skomra:
Hi Peter, you were right I need to 'pip install' as root. I was able to see the tests pass on my the clean system. I wasn't able to figure out what was wrong with the setup on the original system. Thanks for you help with this.

Reviewed-by: Aaron Skomra <aaron.skomra@wacom.com>

Issue #272, opened 2022-06-08
"ninja -C builddir test" returns SyntaxError


Pinglinux:
@whot When running "ninja -C builddir test" with the latest code, I got:
```
--------cut---
1/3 xsetwacom-tests                         OK       0.01 s 
2/3 wacom-tests                             OK       0.04 s 
3/3 pytest                                  FAIL     0.37 s (exit status 4)

Ok:                    2
Expected Fail:         0
Fail:                  1
----------cut---cut---
3/3 pytest                                  FAIL     0.37 s (exit status 4)

--- command ---
01:18:26 GI_TYPELIB_PATH='/home/ping/wacom/xf86-input-wacom/builddir' LD_LIBRARY_PATH='/home/ping/wacom/xf86-input-wacom/builddir' /usr/bin/pytest
--- stderr ---
ImportError while loading conftest '/home/ping/wacom/xf86-input-wacom/test/conftest.py'.
  File "/home/ping/wacom/xf86-input-wacom/test/__init__.py", line 52
    product: int = attr.ib()
           ^
SyntaxError: invalid syntax
```

whot:
what python version is this? this looks like it's complaining about the type annotations but those have been a thing since python 3.5

Pinglinux:
It's on Ubuntu 20.04

$ python --version
Python 3.8.10

Checked the messages in builddir/meson-logs/testlog.txt. It has the following extra lines before the error:

--- stderr ---
No type or invalid type specified.
Must be one of stylus, touch, cursor, eraser, or pad
mod_buttons: Invalid button number 32. Insufficient storage
-------

3/3 pytest                                  FAIL     0.47 s (exit status 4)

whot:
Can you attach the `testlog.txt` please? Right now I don't really have anything to go on beyond "this fails"

Pinglinux:
Here you go: [testlog.txt](https://github.com/linuxwacom/xf86-input-wacom/files/8858116/testlog.txt)

whot:
Note the format of the output, it's divided into the various `test()` specified in `meson.build`. The stderr output you copied above is from the previous test (`wacom-tests`) and expected since it tests for this particular case.

The `pytest` error is still just the syntax error - I'm wondering now: did you pip install anything and if so did you do it as root? If not, it's likely that.

Pinglinux:
> did you pip install anything and if so did you do it as root? If not, it's likely that.

[...]

PR #273, opened 2022-06-08
Better test suite errors

whot:
use debug logging by default and raise an import error with a better explanation. Maybe that "fixes" #272, cc @Pinglinux


Pinglinux:
This patch didn't provide me further information to figure out the setup issue. It didn't break my build more either ;). So, your call to merge or drop...

whot:
If it didn't break anything, that's good enough then :)

PR #274, opened 2022-06-08
CI: add a trigger job to build the test suite on gitlab

whot:
This triggers a workflow run in a custom repository that has a KVM setup
to run the tests that require uinput devices. That repo is nothing more
than a gitlab-ci.yml file that pulls our repo here and runs meson test
inside a QEMU VM. This way we can run those tests that require uinput as
well.

See the repo itself here:
https://gitlab.freedesktop.org/whot/xf86-input-wacom-ci/

Unlike the other workflows, this one only runs when pushing, not on a
PR, for two reasons:
- safer this way since we control the repo (and the secrets required are
  only set in the our repo, not a user fork)
- easier this way, running on a PR would require figuring out which
  branch of which user repo to clone and test. Probably all in the
  gitlab CI variables, but I haven't done this bit yet.

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -1,0 +1,29 @@
+on: [ push ]
+
+env:
+  # whot/xf86-input-wacom-ci
+  project: https://gitlab.freedesktop.org/api/v4/projects/16121
+
+jobs:
+  gitlab-testsuite:
+    name: Run test suite on gitlab
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


whot:
Note that it may be easier to just add the `.gitlab-ci.yml` to this repo and mirror the repo on gitlab. Not 100% whether this will trigger the gitlab pipelines though.

Pinglinux:
Sounds like a good idea.

Acked-by: Ping Cheng <ping.cheng@wacom.com>

whot:
Closing this, I just don't have time to pursue this any further

PR #276, opened 2022-06-11
Added vim modelines to python files.

Greenscreener:
This is a fairly simple edit, but I think it would be very useful, if there was a vim modeline in all the python files too, so at least the indentation doesn't break if one has a custom config.


whot:
Thanks! LGTM but you should invest in an editor that supports `.editorconfig` too ;)

PR #278, opened 2022-06-13
gitignore: add __pycache__

whot:
Might as well not accidentally commit this one

Inspired by @Greenscreener  :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,4 @@ core
 config-ver.h
 test-driver
 version
+__pycache__


.gitignore
```diff
@@ -88,3 +88,4 @@ core
 config-ver.h
 test-driver
 version
+__pycache__
```
Pinglinux, line 91: Shouldn't it be <test/__pycache__> instead?

Also, do we ignore builddir/ and doc/?

I see src/.dirstamp and src/x11/.dirstamp in the untracked files list too. Sorry for the noise :)

whot: gitignore by default matches patterns anywhere, so `__pycache__` will match any such directory anywhere. Since there's no good reason to ever commit `__pycache__`, we might as well not limit ourselves to a single directory - even though we only ever expect that have that file.

`.dirstamp` is ... I think from Make? We should probably add that too then, I'll let you file a separate PR for that though :)

Pinglinux: I only see the .dirstamp files under the repo I cloned a while back. I cloned a new one. I don't see the files any more. I don't know why. I think we don't need to worry about them now. So, feel free to merge your patch.

PR #280, opened 2022-06-19
Filter initial pressure

Pinglinux:


src/wcmCommon.c
```diff
@@ -1250,6 +1255,10 @@ rebasePressure(const WacomDevicePtr priv, const WacomDeviceState *ds)
 	else
 		min_pressure = min(priv->minPressure, ds->pressure);
 
+	/* filter pressure lower than pressure threshold */
+	if (min_pressure < min_pressure_threshold)
```
Singh212dab, line 1259: if (min_pressure < min_pressure_threshold || min_pressure = min_pressure_threshold);

Singh212dab (changes requested):

src/wcmCommon.c
```diff
@@ -1843,27 +1852,50 @@ TEST_CASE(test_rebase_pressure)
...
+	WacomCommonRec common = {0};
 	WacomDeviceState ds = {0};
 	int pressure;
 
-	priv.minPressure = 4;
-	ds.pressure = 10;
+	priv.common = &common;
+	common.wcmMaxZ = 2000;
 
 	/* Pressure in out-of-proximity means get new preloaded pressure */
```
whot, line 1862:
```suggestion
	/* Pressure in an out-of-proximity event means get new preloaded pressure */
```

src/wcmCommon.c
```diff
@@ -1843,27 +1852,50 @@ TEST_CASE(test_rebase_pressure)
...
+	/* pre-registered pressure > pressure_threshold */
+	ds.pressure = 100;
+
+	base = priv;
 
+	pressure = rebasePressure(&priv, &ds);
+	assert(pressure == ds.pressure);
 	assert(memcmp(&priv, &base, sizeof(priv)) == 0);
 
 	/* Pressure in-proximity means rebase to new minimum */
```
whot, line 1885:
```suggestion
	/* Pressure while in-proximity means rebase to new minimum */
```

src/wcmCommon.c
```diff
@@ -1843,27 +1852,50 @@ TEST_CASE(test_rebase_pressure)
...
 
-	priv.minPressure = 4;
-	ds.pressure = 10;
+	priv.common = &common;
+	common.wcmMaxZ = 2000;
 
 	/* Pressure in out-of-proximity means get new preloaded pressure */
 	priv.oldState.proximity = 0;
 
+	/* pressure < pressure_threshold */
```
whot, line 1865:
```suggestion
	/* pressure < min_pressure_threshold */
```

src/wcmCommon.c
```diff
@@ -1843,27 +1852,50 @@ TEST_CASE(test_rebase_pressure)
...
 	/* make sure we don't touch priv, not really needed, the compiler should
 	 * honor the consts but... */
 	base = priv;
 
 	pressure = rebasePressure(&priv, &ds);
-	assert(pressure == ds.pressure);
+	assert(pressure > ds.pressure);
+	assert(memcmp(&priv, &base, sizeof(priv)) == 0);
+
+	/* pre-registered pressure > pressure_threshold */
```
whot, line 1876:
```suggestion
	/* pre-loaded pressure > min_pressure_threshold */
```

we use "pre-loaded" in the other comment, should stick with that

src/wcmCommon.c
```diff
@@ -1843,27 +1852,50 @@ TEST_CASE(test_rebase_pressure)
...
 	priv.oldState.proximity = 1;
 
+	priv.minPressure = 20;
+	ds.pressure = 10;
+
+	base = priv;
+
+	pressure = rebasePressure(&priv, &ds);
+	assert(pressure == priv.minPressure);
+	assert(memcmp(&priv, &base, sizeof(priv)) == 0);
```
whot, line 1895: IMO this is wrong, it just happens to work because `minPressure` is also the min pressure threshold (1% of 2000 is 20). Or if that was the intention, it's confusing given the comment in line 1885. What I'd expect here is that a pressure below the *current minimum* is rebased down to a new minimum, but must now also be `>=` the `min_pressure_threshold`.

So same thing, but with `minPressure = 50` (or some other value > 20), `ds.pressure = 10` and then the value check should be that `pressure > ds.pressure` **and** `pressure < priv.minPressure`. Or, simpler, just have `const int min_pressure_threshold = 20;` in this test with a comment that it's based on the `wcmMaxZ = 2000` and then always compare against that. If we ever change the 0.01% in the driver, we can easily update this test too.

whot (commented):
The code itself looks fine, thanks. But the tests - oh my, the were quite confusing. Best to liberally sprinkle comments into the test to explain what we're expecting and why. This is code that hardly anyone looks at, so comments at least explain the intention.

Pinglinux:
> Best to liberally sprinkle comments into the test to explain what we're expecting and why.

Explain my thoughts in one block before getting into @whot's  comments in detail.

Due to the introduction of the filter, we need to add at least 2 new unit test cases to cover the corner cases: min_pressure falls below filtered pressure when already in proximity and min_pressure is less than filtered pressure when first time in proximity. One of your suggestions may require another unit test case to cover.

src/wcmCommon.c
Pinglinux: The value was intentional. I wanted to make sure minPressure never falls below the min_pressure_threshold. The comment at line 1885 can be confusing. But the fact is the new minPressure is rebased, which should be the same as the min_pressure_threshold.

src/wcmCommon.c
```diff
@@ -1843,27 +1852,50 @@ TEST_CASE(test_rebase_pressure)
...
+
+	ds.pressure = 100;
```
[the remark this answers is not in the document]
Pinglinux: rebasepressure is to report "the minimum pressure based on the current minimum pressure and the hardware state". Below the min_pressure_threshold is a hardware state, which means the value should be filtered.  The minimum pressure doesn't imply 0 pressure. So rebasepressure is basically to find the smallest pressure we can based on. The added code didn't change that description. It changed your expectation though.

src/wcmCommon.c
whot: >  I wanted to make sure minPressure never falls below the min_pressure_threshold

yes, that's a good check to make but it should be done as such - with the value chosen as it is you don't actually test that because you can't tell whether `minPressure` never changed or whether it got pushed down to `min_pressure_threshold`. Hence why `minPressure` must be higher than the threshold so you can verify that a) it gets pushed down as it should and b) that it doesn't end up as less than `min_pressure_threshold`.

src/wcmCommon.c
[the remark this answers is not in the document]
whot: Maybe we should rename `rebasePressure` to `calculateMinPressure` or something instead? I don't remember why I picked "rebase" here - maybe it did the actual rebasing in some earlier version and then got stripped down to just returning the minimum?

src/wcmCommon.c
Pinglinux: > `minPressure` must be higher than the threshold so you can verify that a) it gets pushed down as it should and b) that it doesn't end up as less than `min_pressure_threshold`

Do you suggest two unit test cases or just one? I'd think we need two. That's why I said your suggestion require another unit test case.

src/wcmCommon.c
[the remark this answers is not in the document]
Pinglinux: Ok, I think I know why you were expecting min_pressure as 0. Refer to [https://github.com/linuxwacom/xf86-input-wacom/blob/master/src/wcmCommon.c#L1420](url). With the new code, we need to set priv->maxRawPressure = min_pressure_threshold instead of 0, refer to lines 1422 and 1441. So, rebasepressure isn't a bad name since it is used to set the base/minimal pressure for normalizePressure. We just need to be careful that we update all the other relevant code as well.

src/wcmCommon.c
whot: yeah, better to split it up into two test cases

src/wcmCommon.c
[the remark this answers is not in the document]
whot: tbh, my head is spinning now from so many different variables that are all slightly different `wcmMaxZ`, `maxRawPressure`, then we have `detectPressureIssue` separate from `setPressureButton` etc. But yes, I agree, somewhere in this section should accommodate for the `min_pressure` that we have now, but don't ask me where without me spending some more time to figure all this out :)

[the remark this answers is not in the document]
Pinglinux: I am thinking if kernel is a better place to address this issue, which covers everything in userland. What do you think?

Pinglinux:
After investigating the relevant code, we think the case has been covered by setPressureButton.

PR #282, opened 2022-07-19
Don't report error when waiting for serial number

Pinglinux:
For some devices, such as AES pens, serial number is 0 when tool
comes into proximity. That means serial number was not ready. It
is a known state. All we needed is to wait for the non-zero serial
number. Don't report it through W_ERROR.

Signed-off-by: Ping Cheng <ping.cheng@wacom.com>

### DIFF
--- a/src/wcmUSB.c
+++ b/src/wcmUSB.c
@@ -1020,11 +1020,18 @@ static void usbParseMscEvent(WacomDevicePtr priv,
 	}
 	else
 	{
-		/* we don't report serial numbers for some tools but we never report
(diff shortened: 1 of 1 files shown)


Pinglinux:
[libinput-record.txt](https://github.com/linuxwacom/xf86-input-wacom/files/9144982/libinput-record.txt)
[hid-recorder-tip-only.txt](https://github.com/linuxwacom/xf86-input-wacom/files/9144987/hid-recorder-tip-only.txt)

"libinput record" and "hid recorder" showing the serial number = 0 events

whot:
thanks!

PR #287, opened 2022-08-31
gwacom: add support for special runtime options

whot:
This is a hack to make it possible to test options that cannot be set
through the xorg.conf, e.g button actions that are set through the
X-specific properties interface.

Expose a new function for "runtime options", though the actual option
implementation needs to be handled in the gwacom wrapper code - i.e.
only options explicitly supported can work.

The currently supported option is "PanButton" which maps the given
button to the PanScroll action.

See https://github.com/linuxwacom/xf86-input-wacom/pull/222#issuecomment-1193299467

### DIFF
--- a/src/gwacom/wacom-device.c
+++ b/src/gwacom/wacom-device.c
@@ -296,6 +296,21 @@ WacomOptions *wacom_device_get_options(WacomDevice *device)
 	return device->options;
 }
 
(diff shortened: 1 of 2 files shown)


Pinglinux (approved):
If Peter says it is a hack, I go with it ;).

PR #288, opened 2022-09-01
x11: de-duplicate the valuator number assignment

whot:
Let's add a helper function for this so we can't mess this up.

Somewhat related to #222 since this adds new axes.

cc @Pinglinux, @jigpu, @Greenscreener


Pinglinux:
Patch looks good! Thank you @whot.

PR #294, opened 2022-11-04
Assign Wacom specific keys only to Wacom devices

Pinglinux:
Most extra features are only for Wacom 0x56a devices. Make them specific to those devices. While on the page, I also cleaned the code base a bit.


whot (approved):
LGTM, thanks!

Issue #307, opened 2023-05-25
Gimp crash when enabling wacom dynamics


DenisSalem:
Hello

As shown here : 

- [Crash when trying to use Wacom Intuos with brush dynamics](https://gitlab.gnome.org/GNOME/gimp/-/issues/9187)
- [Huion H420 crashes GIMP](https://gitlab.gnome.org/GNOME/gimp/-/issues/9470)
- [Fatal Error when activating Wacom dynamics to either screen or window](https://gitlab.gnome.org/GNOME/gimp/-/issues/9385)

There is an issue with wacom tablet running along with Gimp. On this [subreddit](https://www.reddit.com/r/GIMP/comments/12me5c6/comment/jhd1k8v/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) someone find out a workaroung by downgrading xf86-input-wacom to 1.1.0.

I don't know if the issue come from Gimp or X11 but I believe there is something to investigate. In the meantime it could be helpful to get back on Gentoo (or any distro) the previous version of the driver which I believe has been removed few days ago.

Best regards

thesamesam:
I'll go ahead and restore the older version in Gentoo, thanks!

whot:
There are only 12 patches between 1.1.0 and 1.2.0, so it'd be great if someone could bisect this. Realistically, there are only 4 or 5 commits that could even have an effect, with a rough guess going to fe923e927a8ddf4d2e82ef4757c885b06d47fa03 since it's the most invasive one.

whot:
Nevermind, ended up having a few spare minutes and yes: https://github.com/linuxwacom/xf86-input-wacom/commit/fe923e927a8ddf4d2e82ef4757c885b06d47fa03 is the source of the issue.

Reproducer is easy, works with a raw X server and gimp which crashes on the first interaction with a canvas (Intuos Pro/PTH-660). Now, the main question is what gimp is unhappy about here, gdb with `gimp-2.10.34-1.fc38.x86_64` says
```
Thread 1 "gimp" received signal SIGABRT, Aborted.
0x00007ffff6bb7844 in __pthread_kill_implementation () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff6bb7844 in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007ffff6b66abe in raise () at /lib64/libc.so.6
[...]

PR #309, opened 2023-05-29
Add option SmoothPanscrollingEnabled

whot:
This is a backwards-compatibility option similar to the Pressure2K
option. GDK2 applications are effectively limited to 7 axes
(value of GDK_AXES_LAST in GDK2) and adding smooth panscrolling gives
our device 8 axes total. This can cause issues when the driver, GDK and
the application don't agree on the number of axes in the device.

This is an application/GDK bug but as for the Pressure2K option we
expect there to be applications that cannot be updated easily. To work
around this, provide a driver option to disable this new feature
altogether and effectively return to 6 axes for the pen device.

This is a partial revert of https://github.com/linuxwacom/xf86-input-wacom/commit/fe923e927a8ddf4d2e82ef4757c885b06d47fa03 "Implement smooth panscrolling"
for the implementation. Minor changes though, the implementation now
relies more on local variables than pointers.

This is in response to #307, Gimp crashing with xf86-input-wacom 1.2.0 because we now have 8 axes on the pen.

cc @Greenscreener, @jigpu, @Pinglinux 

--- 
Note: not fully tested yet, just throwing this out as FYI and so there's no duplication of effort.

### DIFF
--- a/man/wacom.man
+++ b/man/wacom.man
@@ -288,6 +288,17 @@ scroll event is generated when using the "pan" action. Smaller values
 will require less distance and be more sensitive. Larger values will
 require more distance and be less sensitive.  Default: 1300 or 2600
 depending on tablet resolution (corresponds to 13 mm of distance).
+.TP 4
+.B Option \fI"SmoothPanscrollingEnabled"\fP \fI"bool"\fP
+Allows to disable smooth panscrolling. Default: true.
+If disabled, panscrolling sends legacy button events instead.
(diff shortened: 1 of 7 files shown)


Pinglinux:
I am ok with this patchset, especially since it fixes the Gimp crash issue. The X driver is for old systems anyway.

whot:
Added a test case for vertical scrolling since that was broken thanks to a typo, fixed now.

PR #311, opened 2023-06-06
release.sh: drop support for modfiles and module sets

whot:
Leftover from the xorg release.sh script, no longer needed here since we only ever release one module at a time.


Pinglinux:
Well, if you say so :).

PR #330, opened 2023-11-30
wcmUSB: fix undefined behavior for bit-shifting 

whot:
This fixes #329 which fixed the undefined behavior but that was caused by the wrong type. cc @jigpu


jigpu:
LGTM. Wish I had thought of it :sweat_smile:

PR #340, opened 2024-03-04
release.sh: update to take a --token value

whot:
Apparently the old way no longer works, so let's update with a copy/paste from the examples.

Intentional switch to use --token because that's how you should be doing releases anyway - with a temporary token that can be revoked after the release is complete (and that also encodes the username).

https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -96,25 +96,32 @@ release_to_github() {
                         "body": %s,
                         "draft": false,
                         "prerelease": false}' "$tar_name" "$tar_name" "$release_descr")
-    create_result=`curl -s --data "$api_json" -u $GH_USERNAME https://api.github.com/repos/$GH_REPO/$PROJECT/releases`
+    create_result=$(curl -s --data "$api_json" \
+        -H "Accept: application/vnd.github+json" \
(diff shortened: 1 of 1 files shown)


whot:
Note to self: the [wiki page](https://github.com/linuxwacom/xf86-input-wacom/wiki/For-Developers%3A-Making-Releases) needs updating after merging this

Pinglinux:
> Note to self: the [wiki page](https://github.com/linuxwacom/xf86-input-wacom/wiki/For-Developers%3A-Making-Releases) needs updating after merging this

I assume the release.sh for input-wacom should be updated to keep all linuxwacom projects in-sync. What about libwacom? Are you using a better approach for libwacom? Should we adapt the other two projects to libwacom instead?

whot:
libwacom has always used this release.sh anyway, so there's no need to sync here.

Pinglinux:
Thank you. Sounds good to me. Feel free to merge the patch. We'll update release.sh for input-wacom.

PR #341, opened 2024-03-11
release script simplifications

whot:
two cleanups and one simplification: we don't need to enforce the extra dot for the module: `release.sh  .`. We only ever use this script on one repo anyway, so let's imply that.


Pinglinux:
Tested-by: Ping Cheng <ping.cheng@wacom.com>

It worked for me. Thank you Peter!

whot:
thanks!

Issue #339, opened 2024-03-03
Remove old detritus from the man page


whot:
Low-hanging fruit if someone finds the time: the initial parts of the wacom man page are effectively a decade out of date. We don't recommed using `Section InputDevice` anymore, disabling hotplugging is so niche it's not worth mentioning, etc.

flying-elephant:
@whot 
Hi,
I'm working on your request currently and I'd like to make sure before proceeding it.
Do you suppose all of the descriptions related to InputDevice section deleted such as all of the options?
I'd like to make sure of it.

whot:
basically: `Section InputDevice` should be `Section InputClass` with the corresponding examples. The whole driver-internal device hotplugging section needs a rework: `inputclass` and input device hotplugging have been the default for 15 or so years now, so we should default to explaining this.

The actual options should be left alone, unless they don't really matter anymore.

PR #345, opened 2024-06-10
man: remove texts related to InputDevice

flying-elephant:
Since xorg fully started to rely on hotplugging over decade, modern tablet devices can be automatically detected. This commit removes "Section InputDevice" related texts, which is used when hotplugging is disabled.

This changed is related to the issue #339.

### DIFF
--- a/man/wacom.man
+++ b/man/wacom.man
@@ -4,10 +4,12 @@
 wacom \- Wacom input driver
 .SH SYNOPSIS
 .nf
-.B "Section \*qInputDevice\*q"
+.B "Section \*qInputClass\*q"
 .BI "  Identifier \*q" idevname \*q
+.B  "  MatchUSBID \*q056a:*\*q"
+.BI "  MatchDevicePath \*q" devpath \*q
 .B  "  Driver \*qwacom\*q"
-.BI "  Option \*qDevice\*q   \*q" devpath \*q
+.B  "  Option \*q...\*q \*q ...\*q"
 \ \ ...
 .B EndSection
 .fi
@@ -26,9 +28,6 @@ and requires the wacom kernel driver being loaded before this driver starts.
 Please check https://github.com/linuxwacom for latest updates of Wacom X
 and kernel drivers.
 .SH DRIVER-INTERNAL DEVICE HOTPLUGGING
-When input device hotplugging in the X server is enabled and no
-.B InputDevice
-section exists for a compatible tablet device and an
 .B InputClass
 section (see xorg.conf.d(5x)) assigns this driver for the device, the
 .B wacom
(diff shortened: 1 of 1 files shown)


man/wacom.man
```diff
@@ -4,10 +4,12 @@
 wacom \- Wacom input driver
 .SH SYNOPSIS
 .nf
-.B "Section \*qInputDevice\*q"
+.B "Section \*qInputClass\*q"
 .BI "  Identifier \*q" idevname \*q
+.B  "  MatchUSBID \*q056a:*\*q"
```
whot, line 9:
```suggestion
.B  "  MatchIsTablet \*qon\*q"
```
is a better approach here, imo

man/wacom.man
```diff
@@ -26,9 +28,6 @@ and requires the wacom kernel driver being loaded before this driver starts.
 Please check https://github.com/linuxwacom for latest updates of Wacom X
 and kernel drivers.
 .SH DRIVER-INTERNAL DEVICE HOTPLUGGING
-When input device hotplugging in the X server is enabled and no
-.B InputDevice
-section exists for a compatible tablet device and an
 .B InputClass
```
whot, line 31: This needs a `The` prefix, otherwise it sounds a bit awkward. (i.e. `The inputclass blah blah`)

man/wacom.man
```diff
@@ -26,9 +28,6 @@ and requires the wacom kernel driver being loaded before this driver starts.
...
 section (see xorg.conf.d(5x)) assigns this driver for the device, the
```
whot, line 32: not related to your PR but while you're editing this anyway: iirc the `5x` is compression related and up to the distribution, this should just be:

```suggestion
section (see xorg.conf.d(5)) assigns this driver for the device, the
```

whot (changes requested):
thanks, a few minor nitpicks, I'd appreciate it if you could address those.

man/wacom.man
flying-elephant: Understood. I updated the commit for this.

man/wacom.man
flying-elephant: I found other "5x"s in the description and replaced them to "5" on the commit.

whot:
thanks, squashed and merged.

for the future - you can squash review comments into the original commit(s) where applicable. github.com steers you to a commit + merge request but that's wrong for most project, it's better to have a clean commit message log. So for things where you do e.g. a typo fix it's best to do an interactive rebase (google it :) and squash it in, followed by a force push to your branch. This is doubly important where there are multiple logical changes in a single MR because the github UI only lets you either squash them all together into one commit (bad) or leave them as disjointed "fix up blah" commits (also bad).

flying-elephant:
I understand. I'll keep it in mind:)
Thanks!

PR #346, opened 2024-06-24
man: add the description for ResetArea option

flying-elephant:
ResetArea is an option already available in xf86-input-wacom. And it is the reciprocal option to Area option.
This change adds a description to let users know that xsetwacom has ResetArea to set a tablet back to its default values.


whot:
merged, thanks

PR #349, opened 2024-08-08
Add support for multiple relative wheels

jigpu:
The X driver currently has support for devices with a single relative wheel (e.g. scroll wheels on pucks or pads). This patch set adds support for devices that have a second relative wheel sending REL_HWHEEL events from the kernel. Such a control will produce horizontal scrolling events by default. The xsetwacom tool has been updated to allow configuring these controls via the traditional "RelWheel[2]<Up|Down>" commands, as well as a new set of "Dial[2]<CCW|CW>" commands that place relative wheels on the pad (e.g. recent Huion / XP-Pen devices).

### DIFF
--- a/src/x11/xf86WacomProperties.c
+++ b/src/x11/xf86WacomProperties.c
@@ -269,10 +269,10 @@ void InitWcmDeviceProperties(WacomDevicePtr priv)
 
 	if (IsPad(priv) || IsCursor(priv))
 	{
-		for (i = 0; i < 6; i++)
+		for (i = 0; i < 8; i++)
 			wcmInitWheelActionProp(priv, i);
 		prop_wheel_buttons = InitWcmAtom(pInfo->dev, WACOM_PROP_WHEELBUTTONS, XA_ATOM, 32,
-						 6, (int*)priv->wheel_action_props);
+						 8, (int*)priv->wheel_action_props);
 	}
 
 	if (IsStylus(priv) || IsEraser(priv)) {
--- a/include/wacom-properties.h
+++ b/include/wacom-properties.h
@@ -42,15 +42,13 @@
 /* CARD32, 1 value */
 #define WACOM_PROP_SERIAL_BIND "Wacom Serial ID binding"
 
-/* 8 bit, 4 values, left up, left down, right up, right down
-   OR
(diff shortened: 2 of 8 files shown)


Pinglinux:
The changes look good and clean to me. Thank you!

src/x11/xf86WacomProperties.c
```diff
@@ -269,10 +269,10 @@ void InitWcmDeviceProperties(WacomDevicePtr priv)
 
 	if (IsPad(priv) || IsCursor(priv))
 	{
-		for (i = 0; i < 6; i++)
+		for (i = 0; i < 8; i++)
 			wcmInitWheelActionProp(priv, i);
 		prop_wheel_buttons = InitWcmAtom(pInfo->dev, WACOM_PROP_WHEELBUTTONS, XA_ATOM, 32,
-						 6, (int*)priv->wheel_action_props);
+						 8, (int*)priv->wheel_action_props);
```
whot, line 275: the property here is extended to size 8 but when it's being set in `wcmSetActionsProperty` we also check for `prop->size != size`. This means any current client that sends 6 values will now fail with `BadValue`. Need to extend that check to check for 6 or 8 values and allow both.

It's a bit niche because there isn't really a reason to ever change this particular property, you'd expect callers to set the already-existing target property instead. Problem is that whoeever *is* setting this property is probably 7 layers deep in custom in-house tooling and won't notice for years.

jigpu: I'll put this request in my queue since it *would* be nice to allow (I had forgotten that you brought this up to me previously) and not too difficult to implement. I'm not too worried about it though: this property was previously extended from length 4 to 6 at the end of 2011 (without any fallout that I'm aware of) and I honestly don't view the length of this property as API (only the meaning of the offsets).

FWIW, existing/old builds of xsetwacom are forwards-compatible with this change -- despite them modifying this property in a specific case -- because they read the property (and its length) before making changes and writing the modified version back to the driver.

whot: right, thanks. the only user of those magic button APIs that I'm aware of is xsetwacom anyway (everyone else scripts xsetwacom) so let's resolve this and leave it on your todo list instead :)

PR #352, opened 2024-09-05
conf: add support for Surface IPTS device

TheSnowfield:
Add IPTS support


whot:
See #353 and expect this snippet to be removed again soon when it's no longer necessary. But meanwhile - thanks, merged

PR #354, opened 2024-09-12
conf: bind this driver to all tablets by default

### LINKED ISSUE #353: Attach ourselves to all tablet devices?
tbh I thought we're already doing this and only in #347 realised we didnt (and now #352): should we just change our default configuration to `MatchIsTablet "on"` and done, i.e. anything that udev thinks is a tablet should be handled by our driver if it is installed.

With xf86-input-libinput we have an alternative tablet driver and that is used by default on most distros. xf86-input-wacom is no longer mandatory and the easiest way to sort any issues is either install it (if missing) or remove it (if not missing). I think the overlap between "device doesn't work with the wacom driver" and "need to have the wacom driver installed" is nil, so where that exists the users can work around with their own config files.

cc @jigpu, @Pinglinux

whot:
Historically we've only bound to Wacom devices and a select few other tablets that were handled by this driver. This required adding those tablets which we forgot for years for e.g. Huion/XP-Pen devices (see commit 4e1d6d61).

At this point we should support pretty much anything that is a tablet so let's bind to those by default.

Exceptions for Wacom touchpad/touchscreens stay in the file and so do some other ones that are too hard to test whether they're handled by them generic MatchIsTablet condition.

Closes #353

### DIFF
--- a/conf/70-wacom.conf
+++ b/conf/70-wacom.conf
@@ -8,7 +8,6 @@
 
 Section "InputClass"
         Identifier "Wacom USB tablet class"
-        MatchUSBID "056a:*"
         MatchDevicePath "/dev/input/event*"
         MatchIsTablet "true"
         Driver "wacom"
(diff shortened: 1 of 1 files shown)


TheSnowfield (approved):
LGTM

PR #358, opened 2024-11-17
meson.build: fix variable name when dealing with systemd-unit-dir

fundawang:
I guess these are typos.


whot:
yup, typo. thanks heaps, merged

PR #361, opened 2024-12-02
Add DeJitter functionality to Wacom driver and update documentation

mika314:
- Implement de-jitter logic to suppress small, erratic movements during click based on defined thresholds.
- Introduce three new properties for de-jittering:
  - `DeJitterEnable`: Enables or disables de-jittering (default: off).
  - `DeJitterThreshold`: Sets the spatial threshold for movement suppression (default: 150, range: 0-1000).
  - `DeJitterTimeThreshold`: Sets the time threshold for movement suppression (default: 100ms, range: 0-10000ms).
- Update the man page to include detailed documentation for the new properties.


jigpu:
This is an interesting request! I'm a bit concerned about the suggested default values: a `DeJitterThreshold` of 150 can translate to quite a few pixels, depending on the tablet and screen (e.g. 150 would translate to ~7px for an Intuos Pro Medium paired with a 24" 1440p screen). I have not tried this patch yet to see how it feels with different tablet/screen combos, but it seems like it would make fine / precise drawing more frustrating. Not impossible, of course (the cursor will always move to the actual pen location after `DeJitterTimeThreshold` elapses), just more work to get the cursor to the desired pixel.

How did you arrive at this value? Is it tuned for your ideal feel (and if so, what tablet and screen resolution are you using)? Based on some measure of actual hand tremor magnitude?

mika314:
Drawing with dejitter enabled is quite a bad experience. On my end, I have a script that monitors which app is in focus and, if it's for example GIMP, it disables dejitter. It also switches between absolute and relative for apps that work badly in absolute mode, such as Parsec and Unreal Engine 5.

~7px sounds right. I was aiming to about that value. I did not use any science, just play around with values until it felt right.

PR #369, opened 2026-01-27
Restored usage of toolserials beyond the first one

viralesveras:
Due to a change in 6bd9b92, having more than one toolserial specified would lead to "already have a device with this type/serial" for every tool beyond the first one listed in ToolSerials.

This reverts the relevant portion of 6bd9b92 so that multiple specified Toolserials work properly again.


whot (approved):
LGTM, thanks (though I've paged a lot of this out by now)

PR #371, opened 2026-02-09
Add missing man pages

### LINKED ISSUE #370: man pages are missing for isdv4-serial-*.c
Hi,
isdv4-serial-debugger.c and isdv4-serial-inputattach.c don't have man pages. Debian policy states they should: https://www.debian.org/doc/debian-policy/ch-docs.html#manual-pages

I've used an LLM to generate man pages based on the source files. I'll provide a PR for those. We will currently patch them into our package.

Br,

Pieter

plenaerts:
Fix #370

### DIFF
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -31,10 +31,17 @@ xsetwacommandir = $(APP_MAN_DIR)
 xsetwacomman_PRE = xsetwacom.man
 xsetwacomman_DATA = $(xsetwacomman_PRE:man=@APP_MAN_SUFFIX@)
 
+isdv4debuggerdir = $(APP_MAN_DIR)
+isdv4debugger_PRE = isdv4-serial-debugger.man
+isdv4debugger_DATA = $(isdv4debugger_PRE:man=@APP_MAN_SUFFIX@)
 
-EXTRA_DIST = @DRIVER_NAME@.man xsetwacom.man
+isdv4inputattachdir = $(APP_MAN_DIR)
+isdv4inputattach_PRE = isdv4-serial-inputattach.man
+isdv4inputattach_DATA = $(isdv4inputattach_PRE:man=@APP_MAN_SUFFIX@)
 
-CLEANFILES = $(driverman_DATA) $(xsetwacomman_DATA)
+EXTRA_DIST = @DRIVER_NAME@.man xsetwacom.man isdv4-serial-debugger.man isdv4-serial-inputattach.man
+
+CLEANFILES = $(driverman_DATA) $(xsetwacomman_DATA) $(isdv4debugger_DATA) $(isdv4inputattach_DATA)
 
 SUFFIXES = .$(DRIVER_MAN_SUFFIX) $(APP_MAN_SUFFIX) .man
 
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -39,8 +39,6 @@ isdv4inputattachdir = $(APP_MAN_DIR)
 isdv4inputattach_PRE = isdv4-serial-inputattach.man
 isdv4inputattach_DATA = $(isdv4inputattach_PRE:man=@APP_MAN_SUFFIX@)
 
-EXTRA_DIST = @DRIVER_NAME@.man xsetwacom.man isdv4-serial-debugger.man isdv4-serial-inputattach.man
-
 CLEANFILES = $(driverman_DATA) $(xsetwacomman_DATA) $(isdv4debugger_DATA) $(isdv4inputattach_DATA)
 
 SUFFIXES = .$(DRIVER_MAN_SUFFIX) $(APP_MAN_SUFFIX) .man
--- a/man/Makefile.am
+++ b/man/Makefile.am
@@ -39,6 +39,8 @@ isdv4inputattachdir = $(APP_MAN_DIR)
 isdv4inputattach_PRE = isdv4-serial-inputattach.man
 isdv4inputattach_DATA = $(isdv4inputattach_PRE:man=@APP_MAN_SUFFIX@)
 
+EXTRA_DIST = @DRIVER_NAME@.man xsetwacom.man isdv4-serial-debugger.man isdv4-serial-inputattach.man
+
 CLEANFILES = $(driverman_DATA) $(xsetwacomman_DATA) $(isdv4debugger_DATA) $(isdv4inputattach_DATA)
 
 SUFFIXES = .$(DRIVER_MAN_SUFFIX) $(APP_MAN_SUFFIX) .man
--- a/man/isdv4-serial-debugger.man
+++ b/man/isdv4-serial-debugger.man
@@ -35,22 +35,6 @@ file (typically /dev/ttyS0 or similar) and any desired options:
 .B isdv4-serial-debugger /dev/ttyS0
 
 This will connect to the tablet, query its capabilities, and display real-time events.
-.SH OUTPUT
-The tool displays tablet information including:
-.TP
-\- Version and capabilities of the tablet
-.TP
-\- Maximum X and Y coordinates
-.TP
-\- Maximum tilt and pressure values
-.TP
-\- Touch sensor information (if available)
-
-During operation, it shows real-time events in the format:
-.TP
-\- Pen events: position, pressure, tilt, and button states
-.TP
-\- Touch events: position, capacity, and finger status
 .SH EXAMPLES
 .TP
 .B isdv4-serial-debugger \-b 19200 /dev/ttyS0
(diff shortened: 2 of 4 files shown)


man/isdv4-serial-debugger.man
```diff
@@ -0,0 +1,67 @@
...
+Display usage information.
+.TP
+.B \-v, \-\-verbose
+Enable verbose output for debugging.
+.TP
+.B \-V, \-\-version
+Display version information and exit.
+.TP
+.B \-b, \-\-baudrate baudrate
+Set the baud rate for the serial connection. Default is 38400.
```
whot, line 27: those three don't have a short version, only the `--long-option`

man/isdv4-serial-debugger.man
```diff
@@ -0,0 +1,67 @@
...
+.TP
+.B isdv4-serial-debugger \-b 19200 /dev/ttyS0
+Connect to tablet at 19200 baud.
+.TP
+.B isdv4-serial-debugger \-v \-\-reset /dev/ttyS1
+Connect with verbose output and reset the tablet first.
+.SH NOTES
+This tool requires direct access to the serial device, which requires
+appropriate permissions.
+.SH "SEE ALSO"
```
whot: Do you need quotes here?

man/isdv4-serial-inputattach.man
```diff
@@ -0,0 +1,73 @@
...
+.SH OPTIONS
+.TP
+.B \-h, \-\-help
+Display usage information.
+.TP
+.B \-v, \-\-verbose
+Enable verbose output for debugging.
+.TP
+.B \-V, \-\-version
+Display version information and exit.
```
whot, line 24: these two don't have the short option, so best to either fix it in the code or remove it from there :)

man/isdv4-serial-inputattach.man
```diff
@@ -0,0 +1,73 @@
...
+.TP
+\- FUJ devices use 19200 baud
+
+If auto-detection fails, the tool will try the alternative baud rate.
+.SH EXAMPLES
+.TP
+.B isdv4-serial-inputattach \-\-baudrate 19200 /dev/ttyS0
+Connect to tablet at 19200 baud.
+.TP
+.B isdv4-serial-inputattach \-v /dev/ttyS1
```
whot:
```suggestion
.B isdv4-serial-inputattach \-\-verbose /dev/ttyS1
```

man/isdv4-serial-inputattach.man
```diff
@@ -0,0 +1,73 @@
...
+1. Opens the specified serial device
+.TP
+2. Detects or uses the specified baud rate (auto-detects based on device ID if not specified)
+.TP
+3. Queries the tablet to determine its capabilities
+.TP
+4. Sets the serial line discipline to N_MOUSE
+.TP
+5. Binds the kernel driver (W8001 serio type)
+.TP
```
whot: was this AI written? either way, I don't think this section is particularly helpful in a man page, anyone who needs the exact details can look up the code.

man/isdv4-serial-debugger.man
```diff
@@ -0,0 +1,67 @@
...
+.TP
+\- Maximum tilt and pressure values
+.TP
+\- Touch sensor information (if available)
+
+During operation, it shows real-time events in the format:
+.TP
+\- Pen events: position, pressure, tilt, and button states
+.TP
+\- Touch events: position, capacity, and finger status
```
whot: you can skip this whole section, it doesn't add anything IMO

man/Makefile.am
```diff
@@ -31,10 +31,17 @@ xsetwacommandir = $(APP_MAN_DIR)
...
 
-EXTRA_DIST = @DRIVER_NAME@.man xsetwacom.man
+isdv4inputattachdir = $(APP_MAN_DIR)
+isdv4inputattach_PRE = isdv4-serial-inputattach.man
+isdv4inputattach_DATA = $(isdv4inputattach_PRE:man=@APP_MAN_SUFFIX@)
 
-CLEANFILES = $(driverman_DATA) $(xsetwacomman_DATA)
+EXTRA_DIST = @DRIVER_NAME@.man xsetwacom.man isdv4-serial-debugger.man isdv4-serial-inputattach.man
+
+CLEANFILES = $(driverman_DATA) $(xsetwacomman_DATA) $(isdv4debugger_DATA) $(isdv4inputattach_DATA)
```
whot, line 44: please add the new files to the `meson.build` file too, thanks.

Also - do we really need to add those to `EXTRA_DIST` and `CLEANFILES`? I though having them as `_DATA` is sufficient for autotools to DTRT

whot (changes requested):
Thanks, a few nitpicks, the biggest issue is the missing meson integration

man/isdv4-serial-inputattach.man
plenaerts: Yes, I mentioned using an LLM in #370 but didn't repeat here.

I agree.

plenaerts:
Thx for your very quick response. I'll look into your pointers when I find the time.

man/Makefile.am
plenaerts: > Also - do we really need to add those to `EXTRA_DIST` and `CLEANFILES`? I though having them as `_DATA` is sufficient for autotools to DTRT

We need the CLEANFILES. I tested make clean without the extra entries and the compiled man pages remained.
Not sure about EXRA_DIST entries. During debian build the man files are still installed without the EXTRA_DIST entries.

man/isdv4-serial-debugger.man
plenaerts: Confusion here...
Each of these three are in short and long version in the usage() function.
They are in the optstring.
-r is not in the optstring but it is in the options array.

I guess best would be to provide and document short and long versions for each, consistently?

man/isdv4-serial-debugger.man
plenaerts: removed

man/isdv4-serial-inputattach.man
plenaerts: removed

man/isdv4-serial-debugger.man
plenaerts: removed

man/isdv4-serial-inputattach.man
plenaerts: optstring contains only h so I've removed the long versions from the man page.

man/Makefile.am
plenaerts: I'd propose to keep it as is.

I'll look into the meson.build later.

man/isdv4-serial-debugger.man
whot: sorry, i thought I had deleted it before submitting, this applied to inputattach, not this one. PLease ignore.

man/Makefile.am
whot: > Not sure about EXRA_DIST entries. During debian build the man files are still installed without the EXTRA_DIST entries.

EXTRA_DIST simply defines what ends up in the tarball. if `make dist` includes the files then we're good, otherwise we need to add them. We do have a CI job to check this so we'll find out once the CI runs :)

whot:
oh, and once this is ready please squash the commits together, thanks.

man/Makefile.am
plenaerts: Does CI use meson or good old makefiles? 🫣

whot: both actually. and it tests whether we can build with meson from a autotools tarball, so that's all bases covered, I think

plenaerts:
I added the meson.build entries for the two new man pages.

I checked, all installs fine without the EXTRA_DIST entries in man/Makefile.am. I've removed that line.

The CLEANFILES is needed or make clean doesn't cleanup the compiled TROFF files.

> oh, and once this is ready please squash the commits together, thanks.

I think you can easily squash when you merge. If not I'll try it on my side, not sure what that will do to our conversation as commits will disappear...

Anything else from my side in the way of merging?

whot:
> I checked, all installs fine without the EXTRA_DIST entries in man/Makefile.am. I've removed that line.

fwiw `EXTRA_DIST` is for generating a tarball, so to test this you need to `make dist`, then install from that generated tarball. Looks like `EXTRA_DIST` is needed after all, sorry about that.

plenaerts:
> fwiw `EXTRA_DIST` is for generating a tarball, so to test this you need to `make dist`, then install from that generated tarball.

fixed and checked that the tarball contains the man pages.

> Looks like `EXTRA_DIST` is needed after all, sorry about that.

you mentionned it above: https://github.com/linuxwacom/xf86-input-wacom/pull/371#discussion_r2791210029. My bad.

whot (approved):
thanks!

PR #374, opened 2026-03-05
Allow for serial numbers higher than INT_MAX

### LINKED ISSUE #373: Tool serial handling is broken for high serial numbers
I have a pen here with the serial number 0x99800b93 which is unfortunately just past `INT_MAX`, resulting in a parsed serial number of `-1719661677` within the wacom driver. This serial will fail a number of checks in the form `if (serial > -1)`, the immediate effect I noticed (while trying to reproduce #369) is that the tool cannot be assigned its own device with the `ToolSerials` option.

whot:
Our serials are 32-bit unsigned integers except in wcmOptionDupConvert where they were converted to an int. Together with the (serial > -1) comparison this caused valid device serial numbers in the INT_MAX..UINT_MAX - 1 range to be ignored.

Fixes: 3287bf8dd89f ("Store serial numbers as unsigned values")

Closes #373

### DIFF
--- a/src/x11/xf86Wacom.c
+++ b/src/x11/xf86Wacom.c
@@ -235,7 +235,7 @@ void wcmTimerSet(WacomTimerPtr timer, uint32_t millis, WacomTimerCallback func, 
  * @param type Tool type (cursor, eraser, etc.)
  * @param serial Serial number this device should be bound to (-1 for "any")
  */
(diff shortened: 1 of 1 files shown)


Pinglinux:
The patch looks good to me. Thank you, Peter!

